PR #2, opened 2016-09-12
Update README.md

fat763:
Fix links in README.md


lmakarov:
Links are fine and should stay relative, so we do not have static linking to the develop branch.

PR #9, opened 2016-10-07
Default IP 192.168.64.100

lmakarov:
Also no IP switching, network interface fix (kill the interface when VM is removed).


achekulaev:
LGTM

PR #10, opened 2016-10-07
Fix/optimize Windows network settings for file sharing. Fixes #3

### LINKED ISSUE #3: Windows 7: Remote I/O error
The issues is described in https://github.com/blinkreaction/boot2docker-vagrant/issues/89

We should automate this fix in `fin install prerequisites`.

lmakarov:
Fixes https://github.com/docksal/docksal/issues/3

The fix is applied during `fin install tools` on Windows and can also be initiated as a standalone command: `fin fix-smb`.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-FIN_VERSION=0.2.0
+FIN_VERSION=0.2.1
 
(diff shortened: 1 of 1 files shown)


achekulaev:
LGTM

PR #43, opened 2016-11-22
Docs/updates

david-hernandez:


docs/custom-commands.md
```diff
@@ -60,10 +55,10 @@ fin looks for lines starting with `##` for command documentation.
 ## 		--force		Try really hard
 ```
 
-fin will output first line of custom command documentation as a short decription in `fin help`.  
+Fin will output first line of custom command documentation as a short decription in `fin help`.  
 Rest of lines will be available as advanced help via `fin help command_name`
 
-See example of command documentation in [phpcs command](../examples/.docksal/commands/phpcs)
+See example of command documentation in [phpcs command](https://github.com/docksal/docksal/blob/develop/examples/.docksal/commands/phpcs)
```
lmakarov: We should not be using absolute URLs, since then branch name gets hardcoded.

docs/custom-commands.md
```diff
@@ -89,4 +84,4 @@ console.log("Custom NodeJS command!")
 
 ## More examples
 
-[Custom command examples](../examples/.docksal/commands)
+[Custom command examples](https://github.com/docksal/docksal/blob/develop/examples/.docksal/commands)
```
lmakarov: We should not be using absolute URLs, since then branch name gets hardcoded.

docs/custom-commands.md
```diff
@@ -6,15 +6,15 @@ It is possible to extend fin with custom commands (per project).
 
     File `.docksal/commands/updb` (Notice **no extension**. Script name should match command name)
 
-    ```bash
+```bash
```
lmakarov: This should be reverted to what it was. In lists anything in-between has to be indented with 4 spaces.

docs/index.md
```diff
@@ -16,9 +18,9 @@ Please review [system requirements](system-requirements.md) before proceeding wi
 ## Setup
 
 1. [Docksal environment setup](env-setup.md)
-    
-    This is done **one time per host** and should be performed by everyone.
+
```
lmakarov: Fix indentation (revert)

docs/project-setup.md
```diff
@@ -1,27 +1,148 @@
...
-2. Open console (Babun on Windows) to the project directory.
-3. Install a default Docksal configuration file (this downloads the latest `docker-compose.yml` file):
-    
-    ```
-    fin install docksal-config
-    ```
+## Default setup
 
-4. Update `docker-compose.yml` as necessary.
+1 - Create a project directory.
```
lmakarov: Lists and indentation with 4 spaces here and below.

docs/settings.md
```diff
@@ -3,43 +3,64 @@
...
+When the following configuration files are added within a project, they can be used to override default settings:
 
-- `.docksal/etc/php5/php.ini` - PHP settings overrides
-- `.docksal/etc/php5/php-cli.ini` - command line PHP settings overrides
+- `.docksal/etc/php/php.ini` - PHP settings overrides
+- `.docksal/etc/php/php-cli.ini` - command line PHP settings overrides
 - `.docksal/etc/mysql/my.cnf` - MySQL settings overrides
 
-Copy [examples/.docksal](../examples/.docksal) into the `/.docksal` folder in your project repo and modify as necessary.
+Copy [examples/.docksal/etc](https://github.com/docksal/docksal/tree/develop/examples/.docksal/etc) into the `/.docksal` folder in your project repo and modify as necessary.
```
lmakarov: Should not be using absolute URLs

docs/settings.md
```diff
@@ -3,43 +3,64 @@
...
+
+```
+
+See list of available tags below and on [Docker Hub](https://hub.docker.com/r/docksal/cli/tags/).
+
+3 - run `fin up`. This will update the project's configuration.
 
 Available PHP versions:
 
 - `5.6` (`image: docksal/cli:stable`) - default
```
lmakarov: cli images which should be used going forward:
- docksal/cli:1.0-php5
- docksal/cli:1.0-php7

david-hernandez:
The thing about the urls is that mkdocs does not like the relative urls, because those links are pointing to files within the project, not within the docs. That works when viewing on github because the whole project is the site. I'll see if there is another way to do it.

Similar issue with the indentation. mkdocs seems to render some of them differently. I'll double check. One thing I'm doing is checking using the markdown preview in php storm and assuming that will look ok once on github. The local server I use for mkdocs I assume will look the same one up on read the docs.

docs/custom-commands.md
david-hernandez: Oh I see, because it falls under (1.) The problem is mkdocs does not handle the code snippet within the list. It gets rendered incorrectly.

david-hernandez: This is what happens. Attached screenshot.

screen shot 2016-11-23 at 11 38 42 am

docs/index.md
david-hernandez: It's an issue with the numbering again. I added some custom css that will hopefully fix that.

PR #47, opened 2016-12-01
Add new command: "fin create-site"

MKorostoff:
This pull request adds a wizard to create a new drupal site from scratch with a single command.  It was inspired by the [`web-init`](https://github.com/docksal/drupal7/blob/master/.docksal/commands/web-init) utility available in the D7 sample project.

## Screenshot:
new

## Video Demonstration
[fin create-site demo](https://www.youtube.com/watch?v=-zrt02Uc6y4)
https://www.youtube.com/watch?v=-zrt02Uc6y4


bin/fin
```diff
@@ -2855,6 +2856,54 @@ config_generate ()
...
+
+	# Todo: is git clone really the best way to do this?  What about
+	# users who don't have a github account, will this work for them?
+	echo -e "${green} Cloning repository...${NC}"
+	git clone -b master https://github.com/docksal/drupal${newSiteVersion}.git ${projectName}
+	cd ${projectName}/.docksal
+
+	# We have to do this sed command twice, to account for lines with more than
+	# one occurence of "drupalX.docksal"
+	sed -i.bak "s/drupal${newSiteVersion}/${projectName}/" docksal.env && rm docksal.env.bak > /dev/null 2>&1
```
lmakarov: `sed -i'' -e` should be a better variation that is also cross-platform.

http://stackoverflow.com/questions/4247068/sed-command-with-i-option-failing-on-mac-but-works-on-linux/14813278#14813278

achekulaev:
Thanks for this. We will have to rework it before merging in to support wordpress as Docksal is not Drupal-bound in general. It may take some time. But as I understand you wanted this for yourself included.

So meanwhile you can still use this functionality if you make an custom command out of this.  Not many know but custom commands can be not only project specific (like init usually is) but it can also be global. 
https://github.com/docksal/docksal/blob/develop/docs/custom-commands.md#global-custom-commands

Actually I'm pretty sure we will end up making it a custom command as well.

bin/fin
MKorostoff: I'd be happy to change it, but this syntax was literally copy/pasted from a thread about this exact issue http://stackoverflow.com/a/22084103/1209832

MKorostoff: OK, I just spun up a linux box to test this (GNU sed 4.2.2 on Ubuntu 14.04) and it worked just fine.  This syntax has now been tested on both Mac OSX and Linux, so I think it's safe.

MKorostoff:
@achekulaev @lmakarov thank you for the speedy code review and replies 👍.  Here are my replies to the issues raised above:

# Need to Support Wordpress
I have pushed up a new commit (057201b) to support wordpress installation, as you requested.  Here is a look at the new version of the wizard:

new

# `sed` version compatibility
As @lmakarov [noted above](#discussion_r90564345), the syntax for `sed` commands varies between Mac and Linux.  This was something I specifically accounted for when authoring the code originally.  I have now tested it on both Mac and Linux and it worked in both cases.

# Command naming
The name I used for this command was `fin create-site`.  @lmakarov suggested changing it to `fin demo` or `fin demo-wizard`.  I worry that this new proposed name implies that the command should only be used for experimentation and learning.  That's not really how I intended it to be used.  I intended it to be used by folks who are starting up new projects.  I did not intend this to create demo sites which are later disposed.  That said, if you would still prefer the name `fin demo` I will change it as requested.

# Core command vs. Custom command
@achekulaev noted above that we'll probably end up making this a custom command.  I'm not really clear on the meaning of this.  I'm happy to revise this in any way you see fit, but I don't know how I would contribute a custom command.  Of course, I can drop a custom command in my personal `~/.docksal/commands` directory, but how would I contribute this back to the community?

# Build failing
I notice my most recent commit failed on Travis.  I'm not sure from looking at the Travis output what's actually causing the failure, but I suspect the issue is unrelated to this code.  Would it be possible to replay this build?

bin/fin
```diff
@@ -793,6 +793,7 @@ show_help ()
 	printh "status (ps) [-a (--all)]" "Show project services status (-a shows all running containers)"
 	printh "reset [service]" "Recreate project services and containers (${yellow}fin help reset${NC})"
 	printh "rm [service]" "Remove project services (${yellow}fin help rm${NC})"
+	printh "create-site" "Download and install Drupal or Wordpress."
```
achekulaev: Could you move it to the line right above the help for `drush [command]`?
```
...
printh "create-site" "Download and install Drupal or Wordpress."
printh "drush [command]" "Execute Drush command (Drupal)"
...
```

MKorostoff: Sure :-)

I have pushed a new commit implementing this change.

bin/fin
```diff
@@ -2855,6 +2856,68 @@ config_generate ()
...
+	done
+	echo ''
+
+	echo -e "What would you like to install?"
+	echo -e "${green}[1]${NC} Drupal 7"
+	echo -e "${green}[2]${NC} Drupal 8"
+	echo -e "${green}[3]${NC} Wordpress"
+	echo -e ""
+	while true; do
+		read -p ": " newSiteVersion
```
achekulaev: Could we rename `newSiteVersion` to `choice`

local choice
read -p ": " choice

bin/fin
```diff
@@ -2855,6 +2856,68 @@ config_generate ()
...
+		case ${newSiteVersion} in
+			1 )
+				newHumanReadableVersion="Drupal 7"
+				newSiteRepoName="drupal7"
+				break
+				;;
+			2 )
+				newHumanReadableVersion="Drupal 8"
+				newSiteRepoName="drupal8"
```
achekulaev: Our experience has shown that URLs should be clear and stored in one place along with all the constants or it will be hard to support them.
Could you move site repo names to constants at the top of the fin like this?

```
REPO_DRUPAL7="https://github.com/docksal/drupal7.git"
REPO_DRUPAL8="https://github.com/docksal/drupal8.git"
REPO_WORDPRESS="https://github.com/docksal/wordpress.git"
```

and then use them like this

```
1)
  target_cms="Drupal 8"
  target_repo="$REPO_DRUPAL7"
  break

...
echo -e "Your site will run ${yellow}${target_cms}${NC}"
git clone -b master ${target_repo} ${project_name}

```

bin/fin
```diff
@@ -2855,6 +2856,68 @@ config_generate ()
...
+
+	# Todo: is git clone really the best way to do this?  What about
+	# users who don't have a github account, will this work for them?
+	echo -e "${green} Cloning repository...${NC}"
+	git clone -b master https://github.com/docksal/${newSiteRepoName}.git ${projectName}
+	cd ${projectName}/.docksal
+
+	# We have to do this sed command twice, to account for lines with more than
+	# one occurence of "drupalX.docksal"
+	sed -i.bak "s/${newSiteRepoName}/${projectName}/" docksal.env && rm docksal.env.bak > /dev/null 2>&1
```
achekulaev: There's a `g` option to replace all matches in one line
```
sed -i.bak "s/${newSiteRepoName}/${projectName}/g" ...
```

bin/fin
```diff
@@ -2855,6 +2856,68 @@ config_generate ()
 	fi
 }
 
+create_site ()
+{
+	while true; do
+		read -p "Please enter the name of your project (all one word, no spaces or slashes): " projectName
+		#todo, validate this input
```
achekulaev: Please rename projectName to project_name, cleanup and check the input
```
project_name=$(echo "$project_name" | sed -e 's/[^a-zA-Z0-9_-]//')
if [[ "$project_name" == "" ]]; then
  echo-red "Please provide a valid project name"
  exit 1
fi
```

bin/fin
```diff
@@ -2855,6 +2856,68 @@ config_generate ()
...
+	echo -e "What would you like to install?"
+	echo -e "${green}[1]${NC} Drupal 7"
+	echo -e "${green}[2]${NC} Drupal 8"
+	echo -e "${green}[3]${NC} Wordpress"
+	echo -e ""
+	while true; do
+		read -p ": " newSiteVersion
+		case ${newSiteVersion} in
+			1 )
+				newHumanReadableVersion="Drupal 7"
```
achekulaev: And I would like it renamed to `target_cms`. Please see big comment below

bin/fin
achekulaev: Could you also make branch to be `${FIN_BRANCH}`

`git clone -b ${FIN_BRANCH} ${target_repo} ${project_name}`

This is for dev testing purposes.

bin/fin
MKorostoff: yes, I will make this change right away

bin/fin
MKorostoff: yes, I will make this change right away

bin/fin
MKorostoff: I will make the change regarding storing URLs as variables, but I believe the change regarding `FIN_BRANCH` would break this command, because FIN_BRANCH has the value "develop".  None of the sample projects have a branch called "develop"

bin/fin
MKorostoff: this precise syntax didn't work for me, but I've managed to fulfill the spirit of the request like so:

```bash
while true; do
  read -p "Please enter the name of your project (all one word, no spaces or slashes): " project_name
  if [[ "$project_name" =~ [^a-zA-Z0-9_-] ]]; then
    echo-red "Please provide a valid project name (only letters, numbers, dashes, and underscores are permitted)."
    continue
  fi
  break
done
```

MKorostoff:
@achekulaev I have pushed a new commit addressing all of the issues noted in [your recent code review](https://github.com/docksal/docksal/pull/47/files/057201bf6a1d38ab7f37815a3b1afaa19094ff4a)

achekulaev:
Thanks @MKorostoff! You're fast! 
Looks good to me @lmakarov

lmakarov:
Do we want automated test coverage for the command and all options (D7, D8 and Wordpress) or would it be better to introduce separate test suites in the respective sample project repos?

MKorostoff:
This is an interesting question—on the one hand, it seems a little weird to me to have a test in fin core that could be broken by changes to an external repository.

On the other hand, it seems a rather radical change to introduce automated testing to each sample projects just to get coverage on this one command (plus, having an independant test suite per sample project sounds pretty hard to maintain, if you have a goal of adding more CMS'es to docksal).

So I guess I would vote to keep the test(s) for `fin create-site` within docksal core.  This seems to me like it would be much less work in the immediate term, and much easier to maintain in the long term.

PR #63, opened 2016-12-12
Volumes, configuration and docs updates

lmakarov:
- Switched to Docker Compose v1.9 and updated config files to compose file format v2.1
- Enabled project volumes provisioning by default. Experimental, but does not break anything for any existing projects. Used in docksal/drupal8
- `fin generate --static` to generate a standalone project stack config, which will not be updated with a new version of Docksal
- Documentation updates


achekulaev:
We need to update demo projects to config 2.1

lmakarov:
Sample projects updated

PR #64, opened 2016-12-13
Remove containers for non-existing projects

michael-molchanov:


bin/fin
```diff
@@ -672,6 +672,22 @@ _remove_containers ()
 	fi
 }
 
+# Remove containers for removed projects.
+_remove_containers_removed_project_files ()
+{
+	echo-green "Removing containers for removed projects..."
```
achekulaev: Could you remove this message please and replace it with `check_docker_running`

achekulaev:
Looks good to me

lmakarov:
I would like to see a bats test for this use case prior to merging it in.

bin/fin
```diff
@@ -672,6 +672,25 @@ _remove_containers ()
 	fi
 }
 
+# Remove containers for removed projects.
+_remove_containers_removed_project_files ()
```
lmakarov: This function name is not good and difficult to tell what it is about.
Let's change it to `_remove_dangling_containers`

bin/fin
```diff
@@ -705,6 +724,7 @@ _docker_cleanup ()
 
 	_remove_dangling_images
 	_remove_dangling_volumes
+	_remove_containers_removed_project_files
```
lmakarov: Change to `_remove_dangling_containers`

PR #72, opened 2016-12-16
fin config generate changes

achekulaev:
Accept stack as parameter. Possible values: empty value, default (same as empty), acquia
Do not remove .docksal folder if exists. Just overwrite config files.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -2407,7 +2407,9 @@ update ()
 			--format '{{.Label "io.docksal.project-root"}}'
 		)
 		for project in ${running_projects}; do
-			(cd "$project" && fin up)
+			# use updated version when updated
+			[[ ! -f "$FIN_PATH_UPDATED" ]] && FIN_PATH_UPDATED="fin"
+			(cd "$project" && "$FIN_PATH_UPDATED" start)
 		done
 	fi
 
--- a/bin/fin
+++ b/bin/fin
@@ -3092,13 +3092,13 @@ config_generate ()
 		exit 1
 	fi
 
-	if [[ -f .docksal/docksal.yml ]]; then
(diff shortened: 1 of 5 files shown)


bin/fin
```diff
@@ -3088,7 +3092,11 @@ config_generate ()
 {
 	eval $(parse_params "$@")
 	# Allow to defined the default stack file.
-	DOCKSAL_STACK=${DOCKSAL_STACK:-default}
+	if [[ "${ARGV[0]}" != "" ]] && [[ "${ARGV[0]}" =~ ^[a-zA-Z0-9]*$ ]]; then
```
lmakarov: Is `[[ "${ARGV[0]}" =~ ^[a-zA-Z0-9]*$ ]]` necessary here?
There is a [[ -f ... ]] check down below anyway.

bin/fin
```diff
@@ -3097,13 +3105,13 @@ config_generate ()
 	if [[ -f ${yml_file} ]]; then
 		if [[ -d .docksal ]]; then
 			echo-yellow ".docksal directory already exists."
-			_confirm "Remove it and start from scratch?"
-			rm -rf ./.docksal
+			_confirm "Overwrite?"
```
lmakarov: It would probably be better to have `fin config generate` and `fin config update`.
`generate` starts from scratch, while update writes over existing stuff without removing the entire `.docksal` folder. Presumably the only reason we'd have this is to preserve the local yml and env files a user may have already locally.

achekulaev: I think having 2 commands is too much. The only function of this command is to copy a proper file from one dir to another. Instead of remembering this command users can just use their file manager to copy. If there's an additional command to remember which copies without deleting, and another one that copies with deleting I feel that efforts of remembering them will not cover for benefits and it will never be used.

PR #79, opened 2016-12-20
Merge bleeding edge

achekulaev:
Use default database. Make database parameter optional.
Use stdin/stdout for importing/dumping when no file provided
Improve environment setup documentation, make it easier to follow as an instruction for newcomers.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-FIN_VERSION=0.15.0
+FIN_VERSION=0.15.1
 
 # Console colors
 red='\033[0;91m'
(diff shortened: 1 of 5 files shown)


bin/fin
```diff
@@ -2640,8 +2638,8 @@ mysql_import ()
 
 # Dump mysql database
 # @params
-# $1 - database name
-# $2 - file name
+# $1 - file name, if ommitted then stdout
+# --db="drupal" to override database username
 # --db-user="admin" to override mysql username
 # --db-password="otherpass" to override mysql password
 mysql_dump ()
```
lmakarov: Add comments about why -i is necessary - to create and use stdin/stdout streams

bin/fin
```diff
@@ -2589,15 +2591,11 @@ mysql_import ()
...
-	local __database="${ARGV[0]}"
-	local __input="${ARGV[1]}"
-
-	if [[ "${ARGV[0]}" == "" ]] || [[ "${ARGV[1]}" == "" ]]; then
-		show_help_mysql-import
-		exit 1
-	fi
+	local __database="${db:-default}"
+	# important: don't run interactive commands until /dev/fd/0 is used or it's value will be lost
+	local __input="${ARGV[0]:-/dev/fd/0}"
```
lmakarov: Mention what `/dev/fd/0` means and is for

bin/fin
achekulaev: Done

PR #81, opened 2016-12-20
Add bats tests for cleanup process.

michael-molchanov:


achekulaev:
@michael-molchanov please create feature branch in docksal repo and re-raise

PR #92, opened 2016-12-23
Add docksal terminal start script and shortcut.

michael-molchanov:


scripts/docksal.bat
```diff
@@ -0,0 +1,27 @@
+@echo off
```
lmakarov: I don't we need the entire `babun.bat` script here. This script can be as simple as:

```
@echo off
setlocal enableextensions enabledelayedexpansion

set SCRIPT_PATH=%~dp0
set SCRIPT_PATH=%SCRIPT_PATH:\=/%
set BABUN_HOME=%SCRIPT_PATH%
set CHERE_INVOKING=1
set CYGWIN_HOME=%BABUN_HOME%\cygwin

start "" "%CYGWIN_HOME%\bin\bash.exe" --login

```

Notice I switched from zsh.exe to bash.exe. zsh.exe should not be invoked directly with Babun as that results is some settings not being applied - http://take.ms/ZD6Sdr

bin/fin
```diff
@@ -2415,6 +2433,12 @@ update ()
 	fi
 
 	[ $? -eq 0 ] && echo-green "Update finished"
+
+	if is_windows; then
+		echo-yellow "To use Docksal, start the console using the Docksal shortcut on your Desktop."
```
achekulaev: @lmakarov I don't think we need this message here at all. As we discussed let's just silently install shortcut for now. 
And especially we don't need `read -p` here.

lmakarov: Yeah, you're right. It's best to keep this as a semi-hidden feature for now.

PR #103, opened 2017-01-17
Add fin zsh command

fuzzbomb:
The Docksal CLI container has Zsh available, but using it requires opening a bash shell first, like so:

```
breakfast@tiffanys:~/docksal-projects/drupal8$ fin bash
WARNING: No swap limit support
docker@cli:/var/www$ zsh

docker at cli in /var/www (master●)
$ 
```

This PR adds a `fin zsh` command to complement `fin bash`. 

Since Zsh is only installed in the CLI container, `fin zsh` doesn't accept a service name. `fin zsh db` wouldn't work, for example.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -872,7 +872,7 @@ show_help ()
 	printh "rm [service]" "Remove project services (${yellow}fin help rm${NC})"
 
 	echo
-	printh "bash [service]" "Open shell into container. Defaults to ${yellow}cli${NC}"
+	printh "bash [service]" "Open bash shell into container. Defaults to ${yellow}cli${NC}"
 	printh "zsh" "Open zsh shell into ${yellow}cli${NC}"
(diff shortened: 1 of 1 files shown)


achekulaev:
zsh is more of experimental there. I don't think we're ready to expose it. It works slow with git repos of medium and higher size.
We will consider this for future. 

For now as a workaround I propose you @fuzzbomb  creating either a zsh alias or a [global fin command](https://github.com/docksal/docksal/blob/develop/docs/custom-commands.md#global-custom-commands) named `zsh` which would run
```
fin exec zsh
```

fuzzbomb:
Thanks @achekulaev.

I hadn't considered git performance at all. My usual prompt (on my host OS) doesn't include any git information.

achekulaev:
I'll close this for now. We might return to it later.

PR #108, opened 2017-01-24
Pin docker-engine version.

michael-molchanov:
Checked on local with docker-engine 1.13.0 downgrade to 1.12.6 with fin update --tools command

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -499,7 +499,7 @@ check_docker_running ()
 	check_binary_found 'docker'
 	check_binary_found 'docker-compose'
 	if ! is_docker_native && ! is_docker_version; then
-		echo-error "Required Docker version is $REQUIREMENTS_DOCKER or higher"
+		echo-error "Required Docker version is $REQUIREMENTS_DOCKER"
 		echo -e "Run ${yellow}fin update${NC} to update"
 		exit 1
 	fi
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -2144,9 +2144,13 @@ install_tools_ubuntu ()
 			echo "Stopping docker service..."
 			sudo service docker stop 2>/dev/null
 		fi
+
+		# Pin docker-engine version to avoid apt-get upgrade.
+		echo -e "Package: docker-engine\nPin: version ${REQUIREMENTS_DOCKER}*\nPin-Priority: 1001" | sudo tee /etc/apt/preferences.d/docksal.pref >/dev/null
```
achekulaev: If error code is non-zero installation should exit here

achekulaev:
Please also 
1) "Required Docker version is 1.12.6 or higher" - remove `or higher`
2) Check why tests are failimg

bin/fin
achekulaev: Please wrap /etc/apt/preferences.d/docksal.pref filename into double quotes
`sudo tee "/etc/apt/preferences.d/docksal.pref" >/dev/null`

achekulaev:
Looks good to me

PR #141, opened 2017-03-07
Feature/mysql ping

michael-molchanov:
Add mysqladmin ping exec to check that mysql daemon is alive and ready to work (listening on socket).

First use case is to check in init script that DB can be imported.

Second use case is to check that mysql daemon responds to requests on socket.

Third use case is health check on CI server (monitoring).

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-FIN_VERSION=1.1.0
+FIN_VERSION=1.1.1
 
 # Console colors
 red='\033[0;91m'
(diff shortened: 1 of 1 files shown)


lmakarov:
Here's a docker native alternative - health checks:
https://docs.docker.com/compose/compose-file/#healthcheck
https://docs.docker.com/engine/reference/builder/#healthcheck

@michael-molchanov would you be interested to implement that?
This will require changes in the [docksal/service-db](https://github.com/docksal/service-db) MySQL images as in the [stack files](https://github.com/docksal/docksal/tree/develop/stacks)

lmakarov:
So I've testing this with an example heath check stack someone put together. It uses PostgreSQL, but the point is in how the heath check and dependencies between containers are handled within docker-compose - synchronously or not. Thankfully, everything is synchronous and docker-compose will wait until all dependencies are satisfied and exit only then.

This is the example stack I used:
https://github.com/peter-evans/docker-compose-healthcheck 

And here's the video demoing how this works:
http://take.ms/zoQwW

achekulaev:
I recommend implementing this as an addon and propose to docksal/addons

PR #148, opened 2017-03-14
Docksal usage stats

lmakarov:
Minimal OS and version stats

Ping home is done on:
- daily
- `fin --version`
- `fin sysinfo`

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -3003,10 +3003,10 @@ sysinfo ()
 	# os
 	echo "███  OS & BASICS"
 	uname -a
-	if is_linux; then
-		[ -f /etc/os-release ] && cat /etc/os-release
-		[ -f /etc/system-release ] && cat /etc/system-release
-	fi
+	# OS version
+	is_linux && echo $(lsb_release -si) $(lsb_release -sr)
+	is_mac && echo $(sw_vers -productName) $(sw_vers -productVersion)
+	is_windows && echo $(cmd /c ver)
 	# fin version
 	version
 
--- a/bin/fin
+++ b/bin/fin
@@ -187,7 +187,7 @@ fi
 # Generate Docksal instance uuid if not available
 if [[ "$DOCKSAL_UUID" == "" ]]; then
 	export DOCKSAL_UUID=$(uuidgen)
-	echo "DOCKSAL_UUID=$DOCKSAL_UUID" | tee -a "$CONFIG_ENV"
+	echo "DOCKSAL_UUID=$DOCKSAL_UUID" >> "$CONFIG_ENV"
 fi
 
 #---------------------------- Helper functions --------------------------------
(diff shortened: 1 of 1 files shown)


achekulaev:
I believe stats ping calls should be daemonized to prevent console locking if there's internet slowage or unavailability.

bin/fin
```diff
@@ -2009,9 +2009,19 @@ ssh_add () {
 # Reports version stats
 stats_ping ()
 {
-	local os_version="$(uname -s)"
-	[[ $DOCKSAL_STATS_OPTOUT == 0 ]] && \
-		curl -kfsL --data "v=1&tid=${DOCKSAL_STATS_TID}&cid=${DOCKSAL_UUID}&t=event&ec=version&ea=${os_version}&el=${FIN_VERSION}" ${DOCKSAL_STATS_URL}
+	[[ "$DOCKSAL_STATS_OPTOUT" != 0 ]] && return # Don't run if the user opted out
```
achekulaev: We already have SKIP_VBOX_VERSION_CHECK, should we maybe keep alike options named consistently?
Like SKIP_STATS_GATHERING?

bin/fin
```diff
@@ -107,6 +107,12 @@ DEFAULT_MACHINE_VBOX_RAM='1024' #mb
 DEFAULT_MACHINE_VBOX_HDD='50000' #mb
 DEFAULT_MACHINE_DO_SIZE='1gb' # digitalocean default size (512mb, 1gb, 2gb...)
 
+# Stats
+# fin sends a minimal ping with OS and fin version number
+DOCKSAL_STATS_TID='UA-93724315-1'
+DOCKSAL_STATS_URL='http://www.google-analytics.com/collect'
+DOCKSAL_STATS_OPTOUT=${DOCKSAL_STATS_OPTOUT:-0}
```
achekulaev: We already have SKIP_VBOX_VERSION_CHECK, should we maybe keep alike options named consistently?
Like SKIP_STATS_GATHERING?

bin/fin
```diff
@@ -1991,6 +2006,24 @@ ssh_add () {
...
+	[[ "$DOCKSAL_STATS_OPTOUT" != 0 ]] && return # Don't run if the user opted out
+	[[ "$CI" == "true" ]] && return # Don't run in CI
+
+	# We are passing OS version via a custom User-Agent request header, which GA can parse
+	local user_agent
+	is_linux && user_agent="fin/${FIN_VERSION} (Linux $(lsb_release -si)-$(lsb_release -sr))"
+	is_mac && user_agent="fin/${FIN_VERSION} (Macintosh Intel Mac OS X $(sw_vers -productVersion)))"
+	is_windows && user_agent="fin/${FIN_VERSION} (Windows NT $(uname -s | sed 's/CYGWIN_NT-\(.*\)-WOW/\1/'))"
+
+	curl -kfsL \
```
achekulaev: I believe these curl calls should be daemonized to prevent console locking if there's internet slowage or unavailability.
http://stackoverflow.com/a/15944238/1359178

bin/fin
lmakarov: `SKIP_VBOX_VERSION_CHECK` is a one-off variable, while `DOCKSAL_STATS` is a namespace for several variables.

achekulaev: I propose to group a variable that prevents stats curl call with the SKIP_* one instead of DOCKSAL_STATS_* ones.

PR #149, opened 2017-03-15
Merge Edge into Develop

achekulaev:


bin/fin
```diff
@@ -1048,55 +1016,69 @@ show_help_ssh-add ()
...
-	echo -e "${green}Parameters:${NC} one or more service names."
 	echo-green "Examples:"
-	printh "fin rm" "Remove all project services"
-	printh "fin rm web" "Remove web service"
+	printh "fin pl -a" "	Import dump from stdin into default database"
 	echo
 }
 
-show_help_update ()
+how_help_update ()
```
lmakarov: typo

bin/fin
```diff
@@ -3417,18 +3558,34 @@ case "$1" in
 		shift
 		restart
 		;;
-	ps)
-		_ps
+	project)
+		shift
+		case "$1" in
+			status)
```
lmakarov: Since "project" is a global management command, I don't think we should have "status" (which is local, project level) at this level.

bin/fin
```diff
@@ -3417,18 +3558,34 @@ case "$1" in
...
+				echo -e "Unknown command $1. See ${yellow}fin help project${NC}"
+				exit 1
+				;;
+		esac
+		;;
+	status|ps)
+		project_status
 		;;
-	projects|pr)
+	projects|pl)
```
lmakarov: Add a TODO note that this should be removed and "project list" should be used instead.
I still like the shortcut "pl" though, so let's keep that at the top level.

PR #159, opened 2017-04-04
Offline/usb stick use case

lmakarov:
Save/load docker images:
- `fin save <mode>` (--system, --project, --offline, --all)
- `fin load <file>`

Provision Docksal while offline
- Run `fin save --offline` in a project directory
- `fin vm rm`
- `fin vm start --offline`

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -850,9 +850,9 @@ _vhost_proxy_connect ()
 	docker network connect "$network" docksal-vhost-proxy >/dev/null 2>&1
 	if [[ $? == 0 ]]; then
 		echo-green "Connected vhost-proxy to \"${network}\" network."
+		# Run a dummy container to trigger docker-gen to refresh proxy configuration.
+		docker run --rm --entrypoint=echo ${IMAGE_VHOST_PROXY} >/dev/null 2>&1
 	fi
-	# Run a dummy container to trigger docker-gen to refresh proxy configuration.
-	docker run --rm busybox
 }
 
 #------------------------------ Help functions --------------------------------
@@ -1323,6 +1323,7 @@ docker_machine_create ()
 
 	local machine_name="${1:-$DEFAULT_MACHINE_NAME}"
 	local provider="${2:-$DEFAULT_MACHINE_PROVIDER}"
+	local offline="${3:-0}"
 
 	if is_docker_machine_exist "$machine_name"; then
 		echo-error "Docker Machine '$machine_name' already exists"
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -1471,31 +1477,44 @@ docker_machine_start ()
 			docker_machine_mounts
 	else
 		# only auto-create default machine
+		# TODO: Why do we need this?
```
achekulaev: Because there might be a reference to non-existing virtualbox

bin/fin
```diff
@@ -3799,6 +3863,14 @@ case "$1" in
 		shift
 		cleanup $1
 		;;
+	save)
```
achekulaev: Propose:
`fin config images save`

bin/fin
```diff
@@ -3799,6 +3863,14 @@ case "$1" in
...
+		shift
+		docker_save "$@"
+		;;
+	load)
```
achekulaev: Propose: `fin config images load`

bin/fin
```diff
@@ -850,9 +850,9 @@ _vhost_proxy_connect ()
 	docker network connect "$network" docksal-vhost-proxy >/dev/null 2>&1
 	if [[ $? == 0 ]]; then
 		echo-green "Connected vhost-proxy to \"${network}\" network."
+		# Run a dummy container to trigger docker-gen to refresh proxy configuration.
```
achekulaev: Propose: 
# Run a dummy container to trigger docker-gen inside vhost-proxy to generate proxy configuration.

bin/fin
```diff
@@ -1471,31 +1477,44 @@ docker_machine_start ()
...
+		docker_machine_create "$DEFAULT_MACHINE_NAME" "$DEFAULT_MACHINE_PROVIDER" "$mode" && \
 			# Disable IP change for now
 			#docker_machine_change_ip && \
 			docker_machine_env
-
-		if [ $? -eq 0 ]; then
+		[[ $? != 0 ]] && _docker_machine_start_exception
+
+		# Load system images from docksal-system-images.tar (in the current directory) when provisioning in offline mode
+		if [[ $offline ]]; then
```
achekulaev: Propose: reverse the if clause so that regular case was first and offline case was second.

bin/fin
```diff
@@ -1334,6 +1335,9 @@ docker_machine_create ()
 		virtualbox)
 			check_vbox_version
 
+			# Use a local copy of boot2docker.iso (in the current directory) when provisioning in offline mode
+			[[ "$offline" ]] && URL_BOOT2DOCKER="boot2docker.iso"
```
achekulaev: I propose that default paths were not in current directory but in `~/.docksal`
`~/.docksal/boot2docker.iso`
`~/.docksal/images.tar`

PR #166, opened 2017-04-07
fin image and fin update --portable

lmakarov:
- Adds `fin image` command for image management
- Removes the DigitalOcean docker machine use case
- Adds `fin update --portable` (WIP). Currently only supports boot2docker.iso and Docksal system images

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -846,9 +846,9 @@ _vhost_proxy_connect ()
 	docker network connect "$network" docksal-vhost-proxy >/dev/null 2>&1
 	if [[ $? == 0 ]]; then
 		echo-green "Connected vhost-proxy to \"${network}\" network."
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -2498,14 +2480,20 @@ update_start_script_shortcut()
...
-# Update system conteainers' images
+# Update system service images
 update_system_images ()
 {
 	check_docker_running
 
-	docker pull "${IMAGE_VHOST_PROXY}"
-	docker pull "${IMAGE_DNS}"
-	docker pull "${IMAGE_SSH_AGENT}"
+	if [[ "$PORTABLE" ]] && [[ -f "docksal-system-images.tar" ]]; then
```
lmakarov: Because `if [[ ! "$PORTABLE" ]] || [[ ! -f "docksal-system-images.tar" ]]; then` looks less straightforward.

PR #215, opened 2017-05-12
DNS and proxy IP binding refactoring

lmakarov:
- Set DOCKSAL_IP and DOCKSAL_HOST_IP based on the environment (support D4M and D4W)
- Bind proxy and DNS to DOCKSAL_IP instead of 0.0.0.0
- Allow overriding DOCKSAL_VHOST_PROXY_IP (should be set to 0.0.0.0 in CI)
- dns service will now use the hosts upstream resolver (no need to override this anymore)
- cleaned up old code


bin/fin
```diff
@@ -1295,7 +1308,7 @@ show_help_hosts ()
 	printh "fin hosts [command]" "" "yellow"
 	echo
 	printh "Uses $DOCKSAL_DEFAULT_IP for IP when VirtualBox is used."
-	printh "Uses 127.0.0.1 for IP when DOCKER_NATIVE is set."
+	printh "Uses $DOCKSAL_D4MAC_IP for IP when DOCKER_NATIVE is set."
```
achekulaev: Why `DOCKSAL_D4MAC_IP`? If there's no difference for Native App then there should not be 2 different vars for mac (`DOCKSAL_D4MAC_IP`) and win (`DOCKSAL_D4WIN_IP`) or we should just hardcode `127.0.0.1` here. 

If there is a difference, then here should be an `if` clause.

bin/fin
```diff
@@ -3815,12 +3786,25 @@ if [[ "$DOCKSAL_UUID" == "" ]]; then
...
+elif is_docker_native && is_windows; then
+	export DOCKSAL_IP=${DOCKSAL_D4WIN_IP}
+	export DOCKSAL_HOST_IP=${DOCKSAL_D4WIN_HOST_IP}
+else
+	export DOCKSAL_IP=${DOCKSAL_DEFAULT_IP}
+	export DOCKSAL_HOST_IP=${DOCKSAL_DEFAULT_HOST_IP}
+fi
+# DOCKSAL_DNS is used for internal name resolution (for *.docksal domains) via docksal-dns service.
+# Points to the endpoint where docksal-dns service is listening.
+export DOCKSAL_DNS=${DOCKSAL_IP}
```
achekulaev: We should export `DOCKSAL_DNS1` here for backwards compatibility.

PR #232, opened 2017-05-25
add custom commands to bash autocomplete

vml-bsummers:
Adds custom commands from project and global to the bash auto completion


achekulaev:
Looks good to me. Will merge soon. Sorry for long response time

PR #264, opened 2017-06-17
SSH agent refactoring. Fixes #223

### LINKED ISSUE #223: Refactor ssh-add to not depend on the host user's $HOME mount
Currently `fin ssh-add` depends on the host user's `$HOME` folder to be mounted in the VM, so that it can access the user's SSH keys and load them into the agent (`docksal-ssh-agent` service).

We can remove this dependency by using `docker cp` command to copy the key(s) from the host into the agent. Alternatively the key(s) can be passed inline as an argument to the container (via `docker exec`).

Once this is done, we will be able to break the dependency on the global `/Users` NFS mount on Mac and conflicts this may be causing with other exports and/or software (e.g. Vagrant).

lmakarov:
Removes dependency on the host's $HOME directory mount.
Keys are copied into the `docksal-ssh-agent` container with `docker cp`, loaded into the ssh-agent, then removed from the container.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -2230,6 +2230,8 @@ ssh_add () {
 	fi
 
 	local ssh_path="$HOME/.ssh"
+	# On Windows docker and docker-compose expect a Windows style path
+	is_windows && ssh_path=$(cygpath -w "$ssh_path")
 
 	# Copy all keys from the host to the ssh-agent container
 	docker cp "$ssh_path" docksal-ssh-agent:/
--- a/bin/fin
+++ b/bin/fin
@@ -2219,26 +2219,35 @@ ssh_add () {
 
 	# Check if ssh-agent container is running
 	local running=$(docker inspect --format="{{ .State.Running }}" docksal-ssh-agent 2>/dev/null)
-	[[ "$running" != "true" ]] && return
+	# Start the ssh-agent if not running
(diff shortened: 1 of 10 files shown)


bin/fin
```diff
@@ -2215,30 +2215,44 @@ project_create ()
...
 	local ssh_path="$HOME/.ssh"
-	local key_path=""
+	# On Windows docker and docker-compose expect a Windows style path
+	is_windows && ssh_path=$(cygpath -w "$ssh_path")
 
-	# When no arguments provided, check if ssh-agent already has at least one identity. If so, stop here.
-	docker exec docksal-ssh-agent ssh-add -l >/dev/null
-	if [[ $1 == "" && $? == 0 ]]; then return; fi
+	# Copy all keys from the host to the ssh-agent container
+	# Only do this when given an argument (key name) or nothing, but not for options (e.g. -l)
```
achekulaev: What about default keys? id_dsa? It is not passed as an argument

achekulaev:
Are you sure default keys are properly handled?

lmakarov:
This was tested on all platforms, with default, non-default and passphrase protected keys.
All keys are copied from the host into the agent, then ssh-add runs and if no key was given as an argument it picks the defaults keys automatically.

PR #277, opened 2017-06-28
Addons functionality

achekulaev:
Main change: 

- Ability to install and remove addons (custom commands) per project or globally from statically defined repo docksal/addons

Side changes: 

- `self` auto-alias to refer current project
- cd to `docroot` when using `fin drush` with alias and alias points to PROJECT_ROOT. e.g. `fin @project drush st` or `fin @self drush st` will work
- `fin vm ram` will also show amount of available ram and swap


achekulaev:
Please review later. Merging so I could continue.

PR #329, opened 2017-08-24
New DB commands and improvements

achekulaev:
- 1.25.0. fin db cli now accepts query as an argument and also accepts --db parameter to specify database to run query against.
- 1.26.0. fin db create and fin db drop commands added

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-FIN_VERSION=1.25.0
+FIN_VERSION=1.26.0
 
 # Console colors
 red='\033[0;91m'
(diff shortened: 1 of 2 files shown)


lmakarov:
I'm getting `Warning: Using a password on the command line interface can be insecure.` when using db commands. Can we suppress it?

```
$ fin db create mydb
Warning: Using a password on the command line interface can be insecure.
Database mydb is created

$ fin db list
Warning: Using a password on the command line interface can be insecure.
information_schema
default
mydb
mysql
performance_schema

$ fin db drop mydb
Warning: Using a password on the command line interface can be insecure.
Database mydb was deleted.
```

achekulaev:
It's not that easy to suppress that message and since it's not related to the changes in PR I suggest to have a separate ticket for that.

PR #353, opened 2017-09-14
1.29.0. Update project images during update + some update refactoring…

### LINKED ISSUE #332: Check and update all docksal images during `fin update`
As we may be pushing hotfix image versions, we need fin to pull them from Docker Hub for all docksal images present on a host. This should happen on `fin update`.

achekulaev:
… (Closes #332)

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -3326,7 +3326,7 @@ update ()
 
 	# [STEP 5] Update project images
 	if [[ ${UPGRADE_IN_PROGRESS} == 1 ]]; then
-		echo-green "[STEP 5/$TOTAL_STEPS] Updating project images"
+		echo-green "[STEP 5/$TOTAL_STEPS] Updating Docksal images"
 		for _image in $(docker images --format '{{.Repository}}:{{.Tag}}' | grep docksal | sort | uniq); do
 			docker pull "$_image"
 		done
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -3279,56 +3285,57 @@ update ()
...
+		fi
 	fi
 
-	# With native mode also do it during install
-	if [[ ${UPGRADE_IN_PROGRESS} != 1 ]] && is_docker_native; then
-		echo-green "Updating system images..."
-		update_system_images
+	# [STEP 5] Update project images
+	if [[ ${UPGRADE_IN_PROGRESS} == 1 ]]; then
+		echo-green "[STEP 5/$TOTAL_STEPS] Updating project images"
```
lmakarov: This should say "Updating Docksal images", since it updates all Docksal images, not just those of a specific project

bin/fin
```diff
@@ -4831,7 +4839,8 @@ case "$1" in
 		elif [[ "$1" == "--tools" ]]; then
 			update_tools
 		# TODO: "--config" deprecated, remove at some point
-		elif [[ "$1" == "--config" ]] || [[ "$1" == "--stacks" ]]; then
+		elif [[ "$1" == "--config" ]] || [[ "$1" == "--stack" ]] || [[ "$1" == "--stacks" ]]; then
```
lmakarov: Since this pulls all stack files, lets keep it as `--stacks` and drop other options now.
This is an internal argument and it is not mentioned in help or docs.

Issue #352, opened 2017-09-13
Feature: env and stack files on DOCKSAL_ENVIRONMENT variable


b-connect:
It would be nice to declare stack and env files depending on a enviroment variable. 

Like ```DOCKSAL_ENVIRONMENT=prod```

Load order
- docksal.env
- docksal.prod.env
- docksal-local.env

b-connect:
For now we add this code in our docksal.env

```
if [ "$DOCKSAL_ENVIRONMENT" != "" ]; then
  DOCKSAL_ENVIRONMENT=local
fi
DOCKSLA_ENV_FILE="$(get_project_path_dc)/.docksal/docksal.${DOCKSAL_ENVIRONMENT}.env"
if [ -f $DOCKSLA_ENV_FILE ]; then
  source $DOCKSLA_ENV_FILE
fi
```

achekulaev:
Fixed in the referenced PR

PR #354, opened 2017-09-14
Add possibility for add env files based on $DOCKSAL_ENVIRONMENT

b-connect:
Pull request for #352

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -4124,7 +4124,7 @@ load_configuration ()
 	local SEPARATOR=':'; is_windows && SEPARATOR=';'
 	local env_file="$(get_project_path_dc)/.docksal/docksal.env"
 	local local_docksal_env_file="$(get_project_path_dc)/.docksal/docksal.${DOCKSAL_ENVIRONMENT}.env"
-	local local_env_file="$(get_project_path_dc)/.docksal/docksal-local.env"
+	local local_env_file="$(get_project_path_dc)/.docksal/local.env"
 	if [[ -f "$env_file" ]]; then
 		ENV_FILE="$env_file"
 		fix_crlf_warning "$env_file"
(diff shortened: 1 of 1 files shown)


achekulaev:
Thanks for the PR! Looks good to me. Great job figuring out how the code is set up in that function @b-connect!

lmakarov:
Also, `DOCKSAL_ENVIRONMENT` variable should be printed in `config_show` (above the `PROJECT_ROOT` will be fine).

b-connect:
If we remove docksal-local.env this is a breaking change.

b-connect:
Also i want to have a local.env file od docksal-local.env file.

Why.

We use docksal also for staging and dev systems. We need a possibility to have many instances of a staging system (ex PR Builds).

I would prefer following solution of a loading order

- docksal.env ( in vcs)
- docksal.${DOCKSAL_ENVIRONMENT}.env (in vcs )
- local.env ( not in vcs )

b-connect:
Added output for config_show

lmakarov:
> If we remove docksal-local.env this is a breaking change.

Not at all. 

> `DOCKSAL_ENVIRONMENT="${DOCKSAL_ENVIRONMENT:-local}"`

`local` is just one of the environments and it is set by default. `docksal-local.env` is expected to be gitignored. Same for `docksal-local.yml`.

Switching to a different environment file  is done via setting/passing the `DOCKSAL_ENVIRONMENT` variable. Those files would be committed into the repo.

The result is pretty much identical to what you are looking for with the least amount of changes to fin:

```
docksal.env ( in vcs)
docksal-${DOCKSAL_ENVIRONMENT}.env (in vcs )
docksal-local.env ( not in vcs )
```

> We use docksal also for staging and dev systems. We need a possibility to have many instances of a staging system (ex PR Builds).

We use Docksal internally for locals and for sandboxes (branch/PR builder) as well. In those environments `docksal-local.env` is generated and holds the necessary CI setting overrides.

b-connect:
So we should remove docksal-local.env at all ?

lmakarov:
If by "remove" you mean removing `docksal-local.env` from the project repo, then yes. It is not supposed to go there. I realized we do not have this covered in the docs (will update), but the sample projects explain it well - https://github.com/docksal/drupal8/blob/master/.docksal/docksal.env#L1-L4

PR #366, opened 2017-09-26
Improve the case when the host has no ssh keys yet. Fixes #365

### LINKED ISSUE #365: Error when installing Docksal on a machine without SSH keys
This error shows up anytime Docksal tries to load SSH keys into the ssh-agent, but the `~/.ssh` folder is not there (the host has no SSH keys).

`GetFileAttributesEx C:\Users\admin\.babun\cygwin\home\admin\.ssh: The system cannot find the file specified.`

During first-time install:

```
...
Resetting network settings...
Resetting Docksal services...
 * proxy
 * dns and resolver for .docksal domain
Enabling automatic *.docksal DNS resolver...
Clearing DNS cache...
 * ssh-agent
GetFileAttributesEx C:\Users\admin\.babun\cygwin\home\admin\.ssh: The system cannot find the file sp                         ecified.
Configuring SMB shares...
Enter your Windows account password:
Adding docker SMB share...
Importing ssh keys...
GetFileAttributesEx C:\Users\admin\.babun\cygwin\home\admin\.ssh: The system cannot find the file specified.
```

When running `fin reset ssh-agent`:

```
Resetting Docksal ssh-agent service...
GetFileAttributesEx C:\Users\admin\.babun\cygwin\home\admin\.ssh: The system cannot find the file specified.
```

lmakarov:
If there are not keys yet on the host:
- offer to generate a default SSH key pair if there are not keys yet on the host
- fix private key permissions right away
- copy the generate public key to clipboard

End user experience:

```
...
Resetting Docksal ssh-agent service...
You don't seem to have any SSH keys in the default location (/c/Users/admin/.babun/cygwin/home/admin/.ssh).
Would you like to generate a default SSH key pair now? [y/n]: y
Generating a default SSH key pair...
Your new public SSH key was copied to clipboard
```

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -2484,9 +2484,8 @@ ssh_add () {
 		echo "Generating a default SSH key pair..."
 		ssh-keygen -t rsa -b 4096 -f ${ssh_path}/id_rsa -q -N ""
 		chmod 600 ${ssh_path}/id_rsa
-		# For whatever reason pbcopy is not available in scripts in Babun, so defining it here.
-		is_windows && pbcopy () { cat >/dev/clipboard; }
-		pbcopy < ${ssh_path}/id_rsa.pub && echo "Your new public SSH key was copied to clipboard"
+		echo "Your new public SSH key:"
+		echo && cat ${ssh_path}/id_rsa.pub && echo
 	fi
 
 	# On Windows docker and docker-compose expect a Windows style path
--- a/bin/fin
+++ b/bin/fin
@@ -2476,12 +2476,25 @@ ssh_add () {
 	fi
 
 	local ssh_path="$HOME/.ssh"
-	# On Windows docker and docker-compose expect a Windows style path
-	is_windows && ssh_path=$(cygpath -w "$ssh_path")
+	# Check that ssh_path exists and is a directory, abort otherwise
+	if [[ ! -d "$ssh_path" ]]; then
+		echo-yellow "You don't seem to have any SSH keys in the default location (${ssh_path})."
+		_confirm "Would you like to generate a default SSH key pair now?" --no-exit || return
+		# Generate an SSH key pair
+		echo "Generating a default SSH key pair..."
+		ssh-keygen -t rsa -b 4096 -f ${ssh_path}/id_rsa -q -N ""
+		chmod 600 ${ssh_path}/id_rsa
+		# For whatever reason pbcopy is not available in scripts in Babun, so defining it here.
+		is_windows && pbcopy () { cat >/dev/clipboard; }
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -2476,12 +2476,23 @@ ssh_add () {
 	fi
 
 	local ssh_path="$HOME/.ssh"
-	# On Windows docker and docker-compose expect a Windows style path
-	is_windows && ssh_path=$(cygpath -w "$ssh_path")
+	# Check that ssh_path exists and is a directory, abort otherwise
+	if [[ ! -d "$ssh_path" ]]; then
+		echo-yellow "You don't seem to have any SSH keys in the default location (${ssh_path})."
+		_confirm "Would you like to generate a default SSH key pair now?"
```
achekulaev: Needs to be:

```bash
_confirm "Would you like to generate a default SSH key pair now?" --no-exit
[[ "$?" != "0" ]] && return
# Generate an SSH key pair
echo "Generating a default SSH key pair..."
```

Otherwise `_confirm` will break the script with `exit 1`.

bin/fin
```diff
@@ -2476,12 +2476,23 @@ ssh_add () {
...
+		# Generate an SSH key pair
+		echo "Generating a default SSH key pair..."
+		ssh-keygen -t rsa -b 4096 -f ${ssh_path}/id_rsa -q -N ""
+		chmod 600 ${ssh_path}/id_rsa
+		cat ${ssh_path}/id_rsa.pub >/dev/clipboard && echo "Your new public SSH key was copied to clipboard"
```
achekulaev: This does not to seem to work x-platform.

```
$ cat aaa >/dev/clipboard
-bash: /dev/clipboard: Operation not permitted
```

bin/fin
achekulaev: Or maybe even simpler

```bash
_confirm "Would you like to generate a default SSH key pair now?" --no-exit || return
```

lmakarov:
@achekulaev, updated per your feedback

achekulaev:
As discussed this is going to be a separate feature rather that a message baked into ssh_add

lmakarov:
Long term plan: have a new subset of commands:

```
fin ssh-keys add [key-name]
fin ssh-keys ls
fin ssh-keys rm [key-name]
fin ssh-keys generate
```

`fin ssh-keys generate` would be the one covered in this PR

lmakarov:
Another use case:

- `fin ssh-keys add new-key --permanent` - writes the key name to `$HOME/.docksal/docksal.env` as  `SECRET_SSH_KEY_new_key="new-key"`. That way, the key will be automatically loaded when the ssh-agent is reset.

lmakarov:
Closing in favor of #811

PR #372, opened 2017-10-14
Set proper db permissions in fin create db. Fixes #371

### LINKED ISSUE #371: fin db create does not set proper permissions
When creating a database with `fin db create <db-name>`, the the default user in the db container does not have access to the database.

lmakarov:
- renamed mysql function to _mysql to avoid confusion with the mysql binary
- fixed typos in variable value assignment
- fixed issues with variables in queries not being evaluated inside the db container (add bash -c and normalized quotes)
- removed backticks in queries as they are usually not necessary and cause issues in bash unless escaped with \\\


lmakarov:
I'm still working on updating tests and will post here when done with that.

bin/fin
```diff
@@ -3598,11 +3598,11 @@ mysql ()
 
 	check_docksal_environment
 	container_id="${COMPOSE_PROJECT_NAME_SAFE}_db_1"
-	__mysql_command=$(docker exec "$container_id" bash -c "echo -u$__dump_user -p$__dump_password")
+	__mysql_command=$(docker exec "$container_id" bash -c "echo -u${__dump_user} -p${__dump_password}")
 	if [[ "$_query" != "" ]]; then
-		${winpty} docker exec -it "$container_id" mysql ${__mysql_command} ${__database} -e "$_query"
+		${winpty} docker exec -it "$container_id" bash -c "mysql ${__mysql_command} ${__database} -e \"${_query}\""
```
achekulaev: Long to explain but no. Please return to using mysql.

bin/fin
```diff
@@ -3598,11 +3598,11 @@ mysql ()
...
 	else
-		${winpty} docker exec -it "$container_id" mysql ${__mysql_command}
+		${winpty} docker exec -it "$container_id" bash -c "mysql ${__mysql_command}"
```
achekulaev: Same here

bin/fin
```diff
@@ -3636,23 +3636,19 @@ mysql_db_create ()
 	local __dump_password="${dbpassword:-\$MYSQL_ROOT_PASSWORD}"
 	local __db="${ARGV[0]}"
 	local __db_user="\$MYSQL_USER"
-	local __db_charset="${dbuser:-utf8}"
-	local __db_collation="${dbuser:-utf8_general_ci}"
+	local __db_charset="${dbcharset:-utf8}"
+	local __db_collation="${dbcollation:-utf8_general_ci}"
 
-	[[ "$__db" == "" ]] && echo-error "Provide a database name to create." && exit 1
+	[[ "${__db}" == "" ]] && echo-error "Provide the database name to create" && exit 1
```
achekulaev: Provide any database name not specific. 
Hence "provide a"

bin/fin
```diff
@@ -3636,23 +3636,19 @@ mysql_db_create ()
...
-	if_failed_error "Database creation has failed"
+	if_failed_error "Database creation failed"
 
-	mysql --db-user="$__dump_user" --db-password="$__dump_password" --db="$__db" \
-		'GRANT ALL PRIVILEGES ON `'${__db}'.*` TO `'${__db_user}'`@"%"'
-
-	if_failed_error "Database creation has failed"
-
-	echo "Database ${__db} is created"
+	echo "Database ${__db} has been created"
```
achekulaev: So you removed "has" from "has failed" but here you use "has been created"? :)

bin/fin
```diff
@@ -3713,41 +3709,41 @@ mysql_import ()
...
 			SELECT
 				DEFAULT_CHARACTER_SET_NAME, DEFAULT_COLLATION_NAME
 				INTO
 					@_charset, @_collation FROM information_schema.schemata
 				WHERE
-					schema_name = '$__database';
-			SET @CREATE_TEMPLATE = 'CREATE DATABASE \\\`{DBNAME}\\\` CHARACTER SET {DBCHARSET} COLLATE {DBCOLLATION};';
-			SET @SQL_SCRIPT = REPLACE(@CREATE_TEMPLATE, '{DBNAME}', '$__database');
+					schema_name = '${__database}';
+			SET @CREATE_TEMPLATE = 'CREATE DATABASE {DBNAME} CHARACTER SET {DBCHARSET} COLLATE {DBCOLLATION};';
```
achekulaev: Please stop "fixing" this. You are doing it wrong. Those backticks ARE needed

bin/fin
```diff
@@ -3713,41 +3709,41 @@ mysql_import ()
...
+			SET @SQL_SCRIPT = REPLACE(@CREATE_TEMPLATE, '{DBNAME}', '${__database}');
 			SET @SQL_SCRIPT = REPLACE(@SQL_SCRIPT, '{DBCHARSET}', @_charset);
 			SET @SQL_SCRIPT = REPLACE(@SQL_SCRIPT, '{DBCOLLATION}', @_collation);
 			PREPARE DBCREATE FROM @SQL_SCRIPT;
-			DROP DATABASE \\\`$__database\\\`;
+			DROP DATABASE ${__database};
```
achekulaev: Again, those backticks are needed.

bin/fin
```diff
@@ -3636,23 +3636,19 @@ mysql_db_create ()
...
 
-	[[ "$__db" == "" ]] && echo-error "Provide a database name to create." && exit 1
+	[[ "${__db}" == "" ]] && echo-error "Provide the database name to create" && exit 1
 
 	# set db to space to avoid database parameter for mysql command because it might be nonexistent
-	mysql --db-user="$__dump_user" --db-password="$__dump_password" --db=" " \
-		'CREATE DATABASE `'${__db}'` CHARACTER SET '${__db_charset}' COLLATE '${__db_collation}''
+	_mysql --db-user="${__dump_user}" --db-password="${__dump_password}" \
+		"CREATE DATABASE ${__db} CHARACTER SET ${__db_charset} COLLATE ${__db_collation};" \
+		"GRANT ALL PRIVILEGES ON ${__db}.* TO ${__db_user};"
```
achekulaev: I do not really understand the change. I deliberately made GRANT a separate command to distinguish creation issue and grant issue.

bin/fin
```diff
@@ -3582,28 +3582,33 @@ run_cli ()
...
-mysql ()
+_mysql ()
 {
 	check_winpty_found
+	check_docksal_environment
 	eval $(parse_params "$@")
 
+	container_id="${COMPOSE_PROJECT_NAME_SAFE}_db_1"
+	# Bring in MYSQL env variables from the container, so that we can use them here
+	container_mysql_vars=$(docker exec "$container_id" env | grep MYSQL)
```
achekulaev: Let's `grep "^MYSQL_"` to minimize possibility of grepping wrong variables.

bin/fin
```diff
@@ -3582,28 +3582,33 @@ run_cli ()
...
+	eval ${container_mysql_vars}
```
achekulaev: I think we should separate out these 4 lines into a function named `_mysql_env` and use it in all mysql functions

Issue #387, opened 2017-11-14
Add ability to manually start and stop the Docksal system containers


frederickjh:
**Feature Request Description**

Currently the docksal-dns, docksal-ssh-agent, docksal-vhost-proxy containers are all started with the restart policy of **always**. There also does not seem to be a `fin` command to stop them. I could only find the `fin reset system` that will stop, reload and restart them.

**Reason Why**
This feature request is to add a way to manually start and stop the Docksal system containers. This frees up system resources on the host machine when not doing web development work.

**Ideas for a possible way this can be done**

1. Add Docksal variable to allow setting the restart policy on the three system containers to **no**(or any of the allowed flags), but default to **always** if it is not set.
2. Add `fin` command to start and stop the system containers. Maybe `fin docksal start` and `fin docksal stop`.

**Describe the results you expected:**

1. Docksal and fin will have commands to allow manually starting and stop the system containers  . . 
2. and also allow setting a restart policy to one of the allowed docker restart flags (no, always, on-failure, or unless-stopped).

**Closing Comments**
Please let me know if this is something that you would consider incorporating into Docksal, as I might be willing to take the time to code and submit a pull request.

If so it would also be helpful if you could suggest a name for the environmental variable that could be placed in ~/.docksal/docksal.env that would set the restart policy for the system containers.

Thanks for your consideration of this feature!
Frederick

lmakarov:
It is important to note, that this would be applicable to Linux users only, since Mac and Windows users use `fin vm start` and `fin vm stop` to start and stop the entire Docksal VM.

We can use `fin docksal` or `fin system` (TBD) as a command namespace to control Docksal system services (and other things). There, `start|stop|reset [system service]` would make sense. The existing `fin reset [system service]` command will have to be refactored.

As for the restart policy. I think, it's fine to add this as an override option for power users. Something like `DOCKSAL_SYSTEM_RESTART=${DOCKSAL_SYSTEM_RESTART:-always}`

@achekulaev your thoughts?

achekulaev:
_frees up system resources on the host machine when not doing web development work_
Resource saving from stopping those containers is close to neglectable. Those containers hardly consume any resources. RAM and CPU footprint is probably equal to a single tab in Chrome, that I'm sure everyone keeps a lot of, even when they are not needed.

I find the feature request to stop Docksal system containers a **very** edge case.

lmakarov:
> I find the feature request to stop Docksal system containers a very edge case.

[...]

PR #388, opened 2017-11-15
#387 Added DOCKSAL_SYSTEM_RESTART Docksal variable that can be used …

frederickjh:
…to set the prefered Docker restart policy for the Docksal system containers. This can be set in ~/.docksal/docksal.env

This is the first part of work needed to finish #387. 

This pull request adds a Docksal variable `DOCKSAL_SYSTEM_RESTART` to allow setting the restart policy on the three system containers to **no**(or any of the allowed flags), but default to **always** if it is not set.

Still to implement for #387 :
2. Add fin command to start and stop the system containers


achekulaev:
As discussed in https://github.com/docksal/docksal/issues/387 the request to be able to stop Docksal system containers was addressed in docksal/docksal/pull/432

PR #392, opened 2017-11-18
Add vhosts command

fjmk:
Sometimes it is nice to see which virtual hosts are registered in docksal proxy, especially when you have added mailhog or phpadmin services.

- First commit adds the vhosts command.
- Second commit lists all vhosts after ```up``` and ```start``` commands.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1087,6 +1087,7 @@ show_help ()
 	printh "run-cli (rc) <command>" "Run a command in a standalone cli container in the current directory"
 	printh "image <command>" "Image management commands: registry, save, load (${yellow}fin help image${NC})"
 	printh "hosts <command>" "Hosts file commands: add, remove, list (${yellow}fin help hosts${NC})"
+	printh "vhosts" "List all virtual *.docksal hosts registered in docksal proxy"
 	printh "sysinfo" "Show system information for bug reporting"
 	printh "diagnose" "Show statistics for troubleshooting and bug reporting"
 	printh "version (v, -v)" "Print fin version. [v, -v] prints short version"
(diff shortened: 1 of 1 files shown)


achekulaev:
@fjmk currently you can see all those with `fin project list --all` or shortly `fin pl -a`. There is a virtual host column there.

fjmk:
There is a difference in the output of both commands when you have extra services delivering a web-service:

```
> fin pl -a                                                                                                                                              project             STATUS              virtual host                        project root
drupal7docksal      Up 12 seconds       drupal7.docksal,*.drupal7.docksal   /home/frans/projects/drupal7-docksal
```

and
```
> fin vhosts                                                                                                                                             master [05178d4] modified untracked

Available websites are:
http://*.drupal7.docksal
http://drupal7.docksal
http://mailhog.drupal7.docksal
http://phpadmin.drupal7.docksal
http://webui.docksal
```

lmakarov:
Maybe we should add this to `fin config`, which would make it project specific.
In this case, we'll have to parse the output from `docker-compose config` and not the nginx config in `docksal/vhost-proxy`.

At the global level `fin vhosts` may also be useful, but as a subcommand (2nd level command) of something else, e.g. `fin docksal vhosts` (see the discussion in #387).

fjmk:
As it is now, we show only URL's, defined in the nginx config. Nothing shows up in a unstared project. 

When we use ```fin docker-compose config``` we see the project URL's even when the project is not started. Maybe this can be confusing?

About the subcommand: let us wait until #387 is realized?

achekulaev:
@fjmk ah I see. Thanks.

lmakarov:
I'm actually going to remove `vhosts` call from `up` and `start` as that does not belong there.
When you work with more than one project, seeing a list of vhosts for ALL projects triggered by a local project command (up/start/etc.) does not make sense.

PR #393, opened 2017-11-18
Host IP alias on Docker for Mac/Win and xdebug adjustments. Fixes #389

### LINKED ISSUE #389: Xdebug is not working with Docker for Mac
Xdebug does not seem to be working with Docker for Mac.

lmakarov:
- 1.31.0 Set host IP alias for Docker for Mac/Win
- xdebug settings adjustment for cli
  - Point xdebug to the host IP
  - Also set `PHP_IDE_CONFIG=serverName=${VIRTUAL_HOST}` - this helps with debugging console PHP applications (as per our own docs).

    These settings, being set via env variables, take precedence over settings in php.ini.
    Long term this will be include in the cli image.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-FIN_VERSION=1.30.2
+FIN_VERSION=1.31.0
 
 # Console colors
 red='\033[0;91m'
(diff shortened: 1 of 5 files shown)


bin/fin
```diff
@@ -104,9 +104,11 @@ DOCKSAL_ENVIRONMENT="${DOCKSAL_ENVIRONMENT:-local}"
 # Network settings
 ## Docker for Mac/Win
 DOCKSAL_NATIVE_IP="192.168.64.100"
+DOCKSAL_NATIVE_HOST_IP="192.168.64.1"
```
achekulaev: What is the difference between DOCKSAL_DEFAULT_HOST_IP and DOCKSAL_NATIVE_HOST_IP?

lmakarov: There is no difference at the moment. There may not be going forward, in which case, we will blend the two. I just don't want to do this just yet.

lmakarov:
ping @achekulaev

PR #399, opened 2017-11-27
Update getting started guide

ptmkenny:
Just started using docksal today, but it looks like "fin init" has been replaced by "fin config generate"

### DIFF
--- a/docs/getting-started/setup.md
+++ b/docs/getting-started/setup.md
@@ -19,7 +19,7 @@ With Docksal starting a complete LAMP stack for Drupal, WordPress or a pure HTML
 
 ```bash
 git clone <sample-project-repo>
-fin init
+fin config generate
 ```
 
 Try one of the example projects:


lmakarov:
`fin init` is a project level custom command. I makes sense only in the context of a project, which is already configured to work with Docksal. In the context of the setup docs it's meant to be used with one of the sample project repos we have:

 - https://github.com/docksal/drupal7
 - https://github.com/docksal/drupal8
 - https://github.com/docksal/wordpress
 - https://github.com/docksal/magento

Try those out.

`fin config generate` is meant to be used to setup a new **custom** stack configuration.
In most cases you can just create a `.docksal` folder in your project root to let Docksal know where the project root is. We call this "zero configuration" setup. You can then start extending it with overrides. Read more about that here:
https://docs.docksal.io/en/master/advanced/stack-config/

Hope that helps.

Issue #384, opened 2017-11-12
Debian does not see Docksal hosts


ArtuGit:
I have clean Docksal after installing on clean Debian Stretch.
I forced to add my host to `/etc/hosts`, because there is "Server DNS address could not be found” without it.

It is not the big problem for me, but some service host does not work in Behat testing (Selenium, Browser?) and I still cannot recognize what DNS Behat needs.

How to befriends Docksal and Debian concerning the automatic recognition of all DNS.

frederickjh:
@ArtuGit I do not fully understand your situation. This may be something that you could easier get interactive help with on the Docksal [gitter chat](https://gitter.im/docksal/community-support).

On Ubuntu I see the DNS server in the Docksal container listed in:
`/etc/resolv.conf`
where I see the line:
`nameserver 192.168.64.100`
This is the IP address that Docksal is configured to use for Docker.
This is added by Docksal. This is the first thing I would check.

achekulaev:
Docksal does not edit resolv.conf directly. A network interface with nameserver 192.168.64.100 is being added. Linux handles the rest. However network handling differs from repo to repo, so on Debian (especially older ones) it might not work.

lmakarov:
Could be related to #321

ArtuGit:
Adding `nameserver 192.168.64.100` as the first line into `/etc/resolv.conf` solves the problem with accessibility Docksal hosts by names.

But I still have the problem:
`fin reset dns`
```
WARNING: No swap limit support
Resetting Docksal DNS service and configuring resolver for .docksal domain...
Enabling automatic *.docksal DNS resolver...
 ERROR:  The upstream DNS server (8.8.8.8) is not reachable.
         You are either offline, 8.8.8.8 is not a valid DNS server or DNS requests to 8.8.8.8 are blocked.
         Once the current operation completes, follow the steps below:
           1. Open ~/.docksal/docksal.env and set DOCKSAL_DNS_UPSTREAM to your local network DNS server
           2. Run fin reset dns to see if it works
```
Is it related to this issue?

lmakarov:
@ArtuGit I believe `/etc/resolv.conf` should not be altered manually - https://wiki.debian.org/resolv.conf. Thanks why we have #321, however, I don't know whether that will work with other Debian flavors (including pure Debian) other than Ubuntu.

Regarding the upstream DNS server issues - what do you get from

```
nslookup google.com
nslookup google.com 8.8.8.8
```

[...]

PR #400, opened 2017-11-27
Use nmcli instead of ifupdown

### LINKED ISSUE #321: Switch to Network Manager (nmcli) for network configuration on Ubuntu
#319 covered a problem when DNS settings were not taking effect via `/etc/network/interfaces` (though the custom Docksal subnet was created as expected).

Using `/etc/network/interfaces` for network configuration on Ubuntu is deprecated.

Network Manager is the current way of managing network connections with a cli tool called `nmcli`
https://www.tecmint.com/configure-network-connections-using-nmcli-tool-in-linux/

Looks like `nmcli` is a standard tool across Debian family distros, so we should have the unofficial compatibility across the Debian family preserved.

ohthehugemanatee:
closes #321 

Switch to nmcli instead of modifying /etc/network/interfaces. This is my first time using nmcli, so please someone check my work!

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -2650,15 +2650,10 @@ configure_resolver_linux () {
 
 	# TODO: split resolver and network configuration into different functions
 	# Adding a subnet for Docksal. Make sure we don't do this twice
-	if ! grep -q "$DOCKSAL_DEFAULT_SUBNET" /etc/network/interfaces; then
+  if ! grep -q 'docksal' `nmcli dev status`; then
 		echo-green "Adding a subnet for Docksal..."
-		cat > /tmp/docksal.ip.addr <<EOF
-	up   ip addr add ${DOCKSAL_DEFAULT_SUBNET} dev lo label lo:docksal
-	down ip addr del ${DOCKSAL_DEFAULT_SUBNET} dev lo label lo:docksal
-	dns-nameservers ${DOCKSAL_DEFAULT_IP}
-EOF
-		sudo sed -i '/iface lo inet loopback/r /tmp/docksal.ip.addr' /etc/network/interfaces
-		rm -f /tmp/docksal.ip.addr
+    nmcli con add con-name lo:docksal ifname docksal0 type bridge ip4 ${DOCKSAL_DEFAULT_SUBNET}
+    nmcli con modify lo:docksal ipv4.dns ${DOCKSAL_DEFAULT_IP}
 		sudo ifdown --force lo 2>/dev/null
 		sudo ifup lo
 		sudo resolvconf -u


achekulaev:
Related #384

lmakarov:
On a clean Ubuntu 17.10 (DigitalOcean) neither `ifup`/`ifdown`, nor `nmcli` are available out of the box:

```
root@ubuntu-512mb-sfo2-01:~# ifup
The program 'ifup' can be found in the following packages:
 * ifupdown
 * ifupdown2
 * netscript-2.4
Try: apt install <selected package>
root@ubuntu-512mb-sfo2-01:~# nmcli
The program 'nmcli' is currently not installed. You can install it by typing:
apt install network-manager
```

We will need to check if `nmcli` is available and do `sudo apt install network-manager` otherwise.

lmakarov:
After installing network-manager it has to be started before `nmcli` can be used:

```
root@ubuntu-512mb-sfo2-01:~# nmcli
Error: NetworkManager is not running.
root@ubuntu-512mb-sfo2-01:~# /etc/init.d/network-manager start
[ ok ] Starting network-manager (via systemctl): network-manager.service.
root@ubuntu-512mb-sfo2-01:~# nmcli dev status
DEVICE  TYPE      STATE      CONNECTION 
eth0    ethernet  unmanaged  --         
lo      loopback  unmanaged  --  
```

bin/fin
```diff
@@ -2650,15 +2650,10 @@ configure_resolver_linux () {
...
+  if ! grep -q 'docksal' `nmcli dev status`; then
 		echo-green "Adding a subnet for Docksal..."
-		cat > /tmp/docksal.ip.addr <<EOF
-	up   ip addr add ${DOCKSAL_DEFAULT_SUBNET} dev lo label lo:docksal
-	down ip addr del ${DOCKSAL_DEFAULT_SUBNET} dev lo label lo:docksal
-	dns-nameservers ${DOCKSAL_DEFAULT_IP}
-EOF
-		sudo sed -i '/iface lo inet loopback/r /tmp/docksal.ip.addr' /etc/network/interfaces
-		rm -f /tmp/docksal.ip.addr
+    nmcli con add con-name lo:docksal ifname docksal0 type bridge ip4 ${DOCKSAL_DEFAULT_SUBNET}
```
lmakarov: I don't think this works as anticipated:

```
root@ubuntu-512mb-sfo2-01:~# nmcli con add con-name lo:docksal ifname docksal0 type bridge ip4 192.168.64.1/24
Connection 'lo:docksal' (79ff1fa9-674d-4ee8-8072-185022c9cc0d) successfully added.
root@ubuntu-512mb-sfo2-01:~# nmcli con modify lo:docksal ipv4.dns 192.168.64.100
```

I see the settings are added in `/etc/network/interfaces`:

```
root@ubuntu-512mb-sfo2-01:~# cat /etc/network/interfaces 
# This file describes the network interfaces available on your system
# and how to activate them. For more information, see interfaces(5).

# The loopback network interface
auto lo
iface lo inet loopback
	up   ip addr add 192.168.64.1/24 dev lo label lo:docksal
	down ip addr del 192.168.64.1/24 dev lo label lo:docksal
	dns-nameservers 192.168.64.100

source /etc/network/interfaces.d/*
```

Unfortunately these do not get applied right away, neither after a reboot, so neither of these IPs can be reached:

```
192.168.64.1
192.168.64.100
```

bin/fin
```diff
@@ -2650,15 +2650,10 @@ configure_resolver_linux () {
...
+    nmcli con modify lo:docksal ipv4.dns ${DOCKSAL_DEFAULT_IP}
 		sudo ifdown --force lo 2>/dev/null
```
lmakarov: `ifup`, `ifdown` and `resolvconf` are not available out of the box. Should be either installed or replaced with an alternative command available out out the box (preferable).

bin/fin
lmakarov: Scratch the part about settings in `/etc/network/interfaces`. Those were added after I ran Docksal installation...

bin/fin
achekulaev: `ifup`/`ifdown` is a very default utility. Here is empty Ubuntu Xenial from Scaleway.

```
root@scw-eafd0d:~# which ifup
/sbin/ifup
```

achekulaev: Ok nevermind. After Xenial ifup is no longer available by default... Thanks Ubuntu!

lmakarov:
Closing this, as it looks like all we need on recent Ubuntu versions ([confirmed](https://github.com/docksal/docksal/issues/384#issuecomment-349012108) with 17.10) is:

```
sudo apt-get install ifupdown resolvconf -y
```

PR #401, opened 2017-11-28
Install ifup/ifdown if they are missing. Relates to #321

achekulaev:
Starting from Ubuntu 17.04 ifup/ifdown are not available out of the box.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -3033,7 +3033,7 @@ install_tools_ubuntu ()
 	fi
 
 	# Install ifup/ifdown if they are missing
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -3032,6 +3032,12 @@ install_tools_ubuntu ()
 		if_failed "Docker Compose installation/upgrade has failed."
 	fi
 
+	# Install ifup/ifdown if they are missing
+	if ! (which ifup >/dev/null 2>&1) || ! (which ifdown >/dev/null 2>&1); then
```
lmakarov: Also need to check `resolvconf`

achekulaev: Why?

PR #404, opened 2017-11-29
1.40.0. Enrich <project> commands subset.

achekulaev:
From now on the default way to launch commands against project is via `fin project` commands subset. E.g. `fin project start` or `fin p start`.

Old shorthands like `fin start` still work for legacy compatibility.

That's how new help will look

```
$ fin
  Docksal Fin v1.40.0 commands:	

  project <command>        	Project commands: start, stop, status and more... (fin help project)
  vm <command>             	VM commands: start, stop, restart and more... (fin help vm)
  db <command>             	DB commands: import, dump, list, cli (fin help db)

  bash [service]           	Open shell into service's container. Defaults to cli
  logs [service]           	Show service logs (e.g. Apache logs, MySQL logs)
  exec <command|file>      	Execute a command or a script in cli
  config [command]         	Show or generate configuration (fin help config)

  drush [command]          	Execute Drush command (Drupal)
  drupal [command]         	Execute Drupal Console command (Drupal 8)
  wp [command]             	Execute WP-CLI command (WordPress)

  addon <command>          	Addons management commands: install, remove (fin help addon)
  ssh-add [-lD] [key]      	Adds ssh private key to the authentication agent (fin help ssh-add)
  alias                    	Manage aliases that allow fin @alias execution (fin help alias)
  cleanup [--hard]         	Remove unused Docker images and projects (saves disk space)
  share                    	Create temporary public url for current project using ngrok
  exec-url <url>           	Download script from URL and run it on host (URL should be public)
  run-cli (rc) <command>   	Run a command in a standalone cli container in the current directory
  image <command>          	Image management commands: registry, save, load (fin help image)
  hosts <command>          	Hosts file commands: add, remove, list (fin help hosts)
  vhosts                   	List all virtual *.docksal hosts registered in docksal proxy
  sysinfo                  	Show system information for bug reporting
  diagnose                 	Show statistics for troubleshooting and bug reporting
  version (v, -v)          	Print fin version. [v, -v] prints short version
  update                   	Update Docksal
```

New help for project command:

```
$ fin help project
Project management
  fin project <command> [params]	

Commands:
  start                    	Start project services (shorthand: fin start)
  up                       	Forces re-read of configuration and start project services (shorthand: fin up)
  stop [-a (--all)]        	Stop project services (-a stops all services on all projects) (shorthand: fin stop)
  status                   	List project services (shorthand: fin ps)
  restart                  	Restart project services (shorthand: fin restart)
  reset                    	Recreate project services and containers (fin help reset) (shorthand: fin reset)
  remove or rm             	Stop project services and remove their containers (fin help remove)
  config                   	Show project configuration
  build                    	Build or rebuild services (alias for 'docker-compose build')
  list [-a (--all)]        	List running Docksal projects (-a to show stopped as well) (shorthand: fin pl)
  create                   	Project wizard to create new Drupal, Wordpress, Magento or empty static project.

Examples:
  fin pl -a                	List all known Docksal projects, including stopped ones
  fin reset db             	Reset only DB service to start with DB from scratch
  fin project create       	Start a new project wizard
```

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -784,7 +784,7 @@ check_project_unique ()
 		if [[ "$_project_name" == "$COMPOSE_PROJECT_NAME_SAFE" ]] && [[ "$_project_root" != "$PROJECT_ROOT" ]]; then
 			echo-error "Another project is already using the name '${COMPOSE_PROJECT_NAME_SAFE}'" \
 				"Change the name of the current project by renaming the project folder (folder name defines the project name and has to be unique)" \
-				"or remove the other project's stack by running ${yellow}fin rm${NC} in ${yellow}$_project_root${NC}"
+				"or remove the other project's stack by running ${yellow}fin project remove${NC} in ${yellow}$_project_root${NC}"
 			exit 1
 		fi
 
@@ -793,7 +793,7 @@ check_project_unique ()
 		if [[ "$_project_vhost" =~ ^$VIRTUAL_HOST ]] && [[ "$_project_root" != "$PROJECT_ROOT" ]]; then
 			echo-error "Another project is already using the virtual hostname '${VIRTUAL_HOST}'" \
 				"Use a different hostname for this project by overriding the VIRTUAL_HOST variable in docksal.env" \
-				"or remove the other project's stack by running ${yellow}fin rm${NC} in ${yellow}$_project_root${NC}"
+				"or remove the other project's stack by running ${yellow}fin project remove${NC} in ${yellow}$_project_root${NC}"
 			exit 1
 		fi
 	done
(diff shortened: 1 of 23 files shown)


bin/fin
```diff
@@ -1291,15 +1283,25 @@ show_help_image() {
 
 show_help_project() {
 	echo-green "Project management"
-	printh "fin project <command>" "" "yellow"
+	printh "fin project <command> [params]" "" "yellow"
 	echo
 	echo-green "Commands:"
-	printh "list (pl) [-a (--all)]" "List running Docksal projects (-a to show stopped as well)"
-	printh "create" "Create and install new Drupal, Wordpress or Magento project."
+	printh "start" "Start project services (shorthand: fin start)"
```
lmakarov: "shortcut" maybe?
https://en.wikipedia.org/wiki/Shorthand

bin/fin
```diff
@@ -4826,6 +4828,19 @@ case "$1" in
 		shift
 		restart "$@"
 		;;
+	reset)
+		shift
+		# Make sure docker daemon is accessible
+		check_docker_running
```
lmakarov: Why is this necessary specifically here, but not with other subcommands?

bin/fin
```diff
@@ -4861,9 +4876,60 @@ case "$1" in
...
+				;;
+			restart)
+				load_configuration
+				shift
+				restart "$@"
+				;;
+			reset)
+				shift
+				# Make sure docker daemon is accessible
+				check_docker_running
```
lmakarov: Why is this necessary specifically here, but not with other subcommands?

bin/fin
achekulaev: It actually should be here for every command because commands should not do it themselves so that we could call one commands from another without re-checking docker and throwing unneeded errors. But I didn't want to re-write half of fin right now

achekulaev:
So about the annoyance. For you, me and any advanced user `fin project up` *IS* annoying. Hence the `fin p` shortcut/[shorthand](https://en.wikipedia.org/wiki/Shorthand)

However while showing Docksal to many users I realized, that for majority of users who do not know `fin` and docksal that much it is not that big of a problem to type somewhat longer commands if they provide clarity. 

That's why I don't think we should use `fin p` in the documentation extensively.

bin/fin
achekulaev: Yeah it's not really a shortcut... It like and abbreviation. I think shorthand kinda describes what is happening here so unless you have hard feelings against this specific word I would stick with it.

bin/fin
lmakarov: @david-hernandez do you have a recommendation as a native speaker? :)

david-hernandez: @lmakarov I dont understand the question. Did you mean to put this here, or the comment above about shortcut/shorthand?

bin/fin
lmakarov: @david-hernandez do you have a recommendation as a native speaker? :)

achekulaev:
I have updated the documentation.

To you question. There is no intention to retire old commands `fin up` etc. At lest not at the moment.

bin/fin
david-hernandez: Neither is perfect, but shorthand is ok. I can't think of a better label at the moment, so its use is totally fine.

david-hernandez: The alternative might be to label it as alias.

david-hernandez:
I agree with Alexei about the longer commands. When I'm learning something new or use something infrequently, I much prefer a nice verbose command. It is also better for scripts. When options are just letters and abbreviations, it is hard to remember what they are. `fin project start` is very clear and easy to remember. `fin p s` is not.

achekulaev:
I'm going to merge this one tomorrow if there are no objections @lmakarov

lmakarov:
@achekulaev let's sync-up tomorrow (Friday) and do an interactive review.

PR #432, opened 2017-12-28
fin system. Fixes #387

### LINKED ISSUE #387: Add ability to manually start and stop the Docksal system containers

lmakarov:
Initial pass on `fin system` group of commands. Fixes #387 

The goal is to have a clear and clean way of turning everything related to Docksal ON or OFF. The only exclusion would be Docker itself, when run natively (Linux, Docker for Mac/Win).

```
$ fin help system
Manage Docksal
  fin system <command> [params]	

Commands:
  reset                    	Reset Docksal
  start                    	Start Docksal
  stop                     	Stop Docksal
  status                   	Check Docksal status

Examples:
  fin system reset         	Reset all Docksal system services and settings
  fin system reset dns     	Reset Docksal DNS service
  fin system reset vhost-proxy	Reset Docksal HTTP/HTTPS reverse proxy service (resolves *.docksal domain names into container IPs)
  fin system reset ssh-agent	Reset Docksal ssh-agent service
```

We may want the `fin vm` group assimilated into `fin system` as well at some point, but maybe not right now.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -2669,7 +2669,7 @@ install_proxy_service ()
 	# PROJECT_INACTIVITY_TIMEOUT - defines the timeout of inactivity after which the project stack will be stopped (e.g. 0.5h)
 	# PROJECT_DANGLING_TIMEOUT - defines the timeout of inactivity after which the project stack and code base will be
 	# entirely wiped out from the host (e.g. 168h). WARNING: use at your own risk!
-	docker run -d --name docksal-vhost-proxy --label "io.docksal.group=system" --restart=always \
+	docker run -d --name docksal-vhost-proxy --label "io.docksal.group=system" --restart=unless-stopped \
 		-p "${DOCKSAL_VHOST_PROXY_IP:-$DOCKSAL_IP}:${DOCKSAL_VHOST_PROXY_PORT_HTTP:-80}":80 -p "${DOCKSAL_VHOST_PROXY_IP:-$DOCKSAL_IP}:${DOCKSAL_VHOST_PROXY_PORT_HTTPS:-443}":443 \
 		-e PROJECT_INACTIVITY_TIMEOUT="${PROJECT_INACTIVITY_TIMEOUT:-0}" \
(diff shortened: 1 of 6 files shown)


bin/fin
```diff
@@ -2542,9 +2556,91 @@ project_create ()
...
+		system_reset
+	fi
+}
+
+# Stops/disables all Docksal things
+# TODO: add projects, network, /etc/exports, etc. here for a complete shutdown of Docksal
+system_stop ()
+{
+	if ! is_linux && ! is_docker_native && ! is_wsl; then
+		echo-green "Stopping Docksal VM..."
```
achekulaev: Shouldn't configure_resolver off be here as well?

lmakarov: Yes, however it's already present in `vm stop`

PR #436, opened 2018-01-04
Add option to clone custom git repository for create-project command

roshkovanv:
Sometimes we need different configs for docksal/drupal so it would be great if we have additional option for  'project create' so we can clone from our repo, not docksal's drupal repo.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -2466,9 +2466,10 @@ project_create ()
 	echo -e "${green}   [6]${NC} Grav CMS"
 	echo -e "${green}   [7]${NC} Gatsby JS"
 	echo -e "${green}   [8]${NC} Laravel"
+	echo -e "${green}   [9]${NC} Custom git repository"
 	echo -e ""
 	while true; do
-		read -p "Enter your choice (1-8): " choice
+		read -p "Enter your choice (1-9): " choice
 		case ${choice} in
 			1)
 				target_cms="Drupal 7"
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -2514,8 +2515,20 @@ project_create ()
 				target_host_name_search_string=""
 				target_repo="$URL_REPO_LARAVEL"
 				;;
+	        9)
```
asgorobets: indentation here will need some love

bin/fin
```diff
@@ -2514,8 +2515,20 @@ project_create ()
...
+				target_host_name_search_string=""
+				target_cms="Custom git repository"
+				while true; do
+					read -p "Please enter repository url: " target_repo
+					if [[ -z "${target_repo// }" ]]; then
+						echo-red "Please provide repository url."
```
asgorobets: Wouldn't this output 'Please provide repository url' twice if you didn't provide a value?

achekulaev:
Hi and thanks for contribution. 

`project create` job is to deliver guaranteed working examples of different web projects. In the case of custom user projects user should check out it first, as usual. 

This could be a good idea for global addon though.

asgorobets:
I think the idea here is to have a boilerplate example of drupal8 or drupal7 repo with all docksal configs, but not hosted on docksal git repos, rather than controlling drupal8 base project boilerplate, give user a way to use his own starter project

Yes, basically cloning a Git boilerplate git repo can do the job, but it won't replace the hostname with project name, so an additional step will be needed

achekulaev:
@asgorobets with custom user projects we can not know what is being checked out and wether it even has `init` command defined. It is a black box for fin. What if `init` of that project contains `rm -rf /`? Users will blame Docksal for that.

The name of the project is the name of the folder that contains it. If you want your project to have name "alex1" just check it out into folder "alex1". The problem solves itself.

@lmakarov please post your input too.

achekulaev:
I think we could have this code included but there needs to be a confirmation after checking out the custom repo but before running `init` from that custom repo. See a confirmation example in `exec_url`.

lmakarov:
I think this is a create feature. The PR is old and outdated, so I created a new one: https://github.com/docksal/docksal/pull/1154

@roshkovanv thanks for submitting this in the first place :)

PR #461, opened 2018-01-31
Document Command Namespacing

### LINKED ISSUE #460: Documenting commands as part of subfolders
**Description**

When writing custom commands whether they are global or project specific you can place the commands in subfolders to organize them but when you try to get a list of those commands they aren't showing up in the list.

Problem/feature description...

1. create folder in `~/.docksal/commands/example`
2. create new command in example folder called `cmdexample`
3. type `fin`
4. Command does not show up in list
5. you can execute this command though using `fin example/cmdexample` though

sean-e-dietrich:
Fixes issue where commands are only found when in the immediate folder but if there are subfolders they don't show up.

This will allow for commands to be placed in subfolders and show up in the list of available commands. The purpose this could be used for is if a team wanted to create a repo of shared commands having them stored in a repo and cloning the repo down.

fixes #460

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1170,13 +1170,14 @@ show_help_list_commands_in_help ()
 		addons_path="$HOME/$DOCKSAL_COMMANDS_PATH"
 	fi
 
-	local addons_list=$(ls "$addons_path" 2>/dev/null | tr "\n" " ")
+	local addons_list=$(find "$addons_path" -type f 2>/dev/null)
 	if [ ! -z "$addons_list" ]; then
 		local scope
-		for cmd_name in $(ls "$addons_path")
+		for cmd_name in $(find "$addons_path" -type f)
 		do
 			local g
 			[[ "$1" == 'global' ]] && g=' [g]'
+			cmd_name=${cmd_name/${addons_path}\//}
 			# command description is lines that start with ##
 			local filename="$addons_path/$cmd_name"
 			local cmd_desc


achekulaev:
Proposed change will find all files under all folders in the `commands` folder. Not all files might be commands however. Other files can include documentation or configuration and we do not want to show all those files under commands section in fin. 

I don't think this approach plays well with current ideology of commands. Ability to execute any file in commands folder is actually an unintentional feature that (for reasons above) I would be more inclined to limit than to endorse.

sean-e-dietrich:
My problem with addons are that you can only execute the addons command directly as opposed to having a command to automatically pull up SequelPro or even a Terminus CLIE wrapper. Those each would need to be their own addons. This is fine and not a huge deal just multiple repositories to maintain for multiple addons.

All my change does is list out commands that can already be executed just aren't being documented when you type `fin help`.

achekulaev:
> My problem with addons are that you can only execute the addons command directly as opposed to having a command to automatically pull up SequelPro or even a Terminus CLIE wrapper

I'm not sure I follow. You can have single addon to do multiple things. Addon has a single entry point (main script), but that script can execute other scripts within addon or even other addons or commands.

> multiple repositories to maintain for multiple addons

Again I'm not sure I follow. Addons, just like custom commands, are per-project things. At least usually.

> All my change does is list out commands

You change will list out everything, not just commands. 

demonodb -bash 2 2018-01-31 18-44-02

sean-e-dietrich:
> I'm not sure I follow. You can have single addon to do multiple things. Addon has a single entry point (main script), but that script can execute other scripts within addon or even other addons or commands. Just to make sure we are on the same page. Addon is like richer command. See example addon in https://github.com/docksal/addons.

This sounds like you are suggesting something similar just different method so if I had an addon `example` then in it is a command to run a subcommand `subcommand` it would be something like

```
$ fin example subcommand [arguments]
```

Is that correct?

If the above is correct then this approach could be similar.

> Your change will list out everything, not just commands.

This currently happens right now but only with commands directly in the command folder are picked up. Example: Just created a blank file in `~/.docksal/commands` called `test`. Then ran `fin help` and the file shows up in my `Custom commands:`.

achekulaev:
> Is that correct?

It is exactly correct. Everything after the addon/command name is passed as a parameter to the addon/command script.

It is up to the developer of course to parse those command line arguments in his/her addon, but otherwise you have the full freedom. 

**Example**

You can have addon `launch`

```
.docksal/addons/launch
  launch
  launch_sqlpro
  launch_terminus_cli
```

your `.docksal/addons/launch/launch` would read

```bash
#!/bin/bash

if [[ "$1" == "sqlpro" ]]; then
  exec "$ADDON_ROOT/launch_sqlpro"
elif [[ "$1" == "terminus" ]]; then
  # remove one argument from arguments list
  shift 
  # exec launch_terminus and pass the rest of arguments to it
  exec "$ADDON_ROOT/launch_terminus" "$@"
else
  echo "Usage: fin launch <sqlpro | terminus>"
  exit 1
fi
```

`.docksal/addons/launch/launch_sqlpro` would read:

```bash
#!/bin/bash

echo "Starting Sequel Pro..."
open -a "Sequel Pro"
```

`.docksal/addons/launch/launch_terminus` would read:

```
#!/bin/bash

echo terminus "$@"
```

Set all files executable, and now you can use it

```
fin launch sqlpro
fin launch terminus auth:login --email=dev@example.com
```

Though I would personally separate out terminus into a separate addon for simplicity of usage, but you can do what is more suitable to you.

frederickjh:
@achekulaev Just a quick clarification question about Docksal **addons** currently they are only fetched from the [docksal/addons](https://github.com/docksal/addons) repo, correct?  I believe this is correct, as in working on a new addon I had to keep copying it in to my test project.
If this is so "private" **addons** not it the [docksal/addons](https://github.com/docksal/addons) github repository would need to be copied or cloned into each project manually. With Docksal **commands** you put them once in `~/.docksal/commands/` and they are available for all projects.
Would it be possible to allow for private addons for @sean-e-dietrich use case ? In other word if  `fin` does not find the **addon** in the github repository it then looks in `~/.docksal/addons/` for the addon. This would allow a developer to copy or clone in an **addon** once so that would be available for all projects.

achekulaev:
@frederickjh global addons already work. Try it out.

frederickjh:
@achekulaev, yes I know that global addons work, as I have been using these. 

The question was whether the code could be easily configured to look in a local location for addons if they were not found in the [docksal/addons](docksal/addons) repository.  This would allow for "private" addons like @sean-e-dietrich was wanting for his company. I had suggested in a subdirectory of the users Docksal configuration, ie. `~/.docksal/addons/` as a possible location for this.

I also figured out that if you put your "private" addon in a projects `.docksal/addons/projectname` folder that fin will see it.

achekulaev:
Initially I wrote the functionality in a way that would be most efficient, which is to download only the files required and I had no idea where the whole addons idea would take us. Just wanted to give users some flexibility to "add" functionality without a need to include it into the core. 
There is no easy way to use private repo at the moment. I need to write a separate case for that which would involve checking out the repo first.

frederickjh:
When you say,
> checking on the repo first.
Do you mean someones private addon repo?

I wasn't even thinking that far. I was just thinking if the `fin addon install` command looked in a local folder after the [docksal/addons](docksal/addons) repository people could clone their private addons repository in the folder then docksal would find them. However being able to add addon repositories is another way to do it.

I guess I got the idea from the Docksal commands. You just drop them it the `~.docksal/commands` folder and they are available.

achekulaev:
I do not really get the use case. If you already have your private addons checked out, why wouldn't you copy an addon you need into `~/.docksal/addons` or `$PROJECT_ROOT/.docksal/addons`?

frederickjh:
I am using the use case of @sean-e-dietrich  in the OP. 
> This will allow for commands to be placed in subfolders and show up in the list of available commands. The purpose this could be used for is if a team wanted to create a repo of shared commands having them stored in a repo and cloning the repo down.

I believe he is saying to clone the repository in to a projects `.docksal/addons/`. This pull request would allow commands to be found at a deeper directory levels than currently is done. So currently you can only place your addon in: `.docksal/addons/myaddon`. This code would allow you to place your addons in `.docksal/addons/myteamsaddonrepository/myaddon`.

Currently when you do `fin addon remove` fin purges the addon directory from the project even for private addons. This means if you want to reinstall it you must copy or clone it again into `.docksal/addons/`. This also goes for any new projects that you want to install your private addon to.

The change I am talking about would allow private Docksal addons to work like Docksal commands currently do, in that you place them once in your global `~.docksal/addons/` folder then you can install them on any project without having to copy or clone them into a project's `.docksal/addons/` directory.

In the use case from the OP this change in the code would save lots of time for team members that have to copy or clone the team addons into each project they work on. With this addition the could just do `fin addon install myteamaddon`.

achekulaev:
I see what you mean. I don't see how to properly implement it right now, but I see the use case now. Except this part.

> allow commands to be found at a deeper directory levels than currently is done

Could you explain where the ask for this comes from? Why is there a desire to make the file structure deeper, which Sean seems to share? Is your automation that sophisticated that in requires namespaces?

sean-e-dietrich:
@achekulaev my specifics would be we create a toolbox for our company which has commands we use across all projects. We put them in a repository and that repository can be cloned locally and then updates could be pulled down. But, I also want to have custom global commands that I use for other projects.

sean-e-dietrich:
I've recently got involved with add ons and starting to understand them a little more but am not a huge fan of having to create a routing command in order to do what I'm looking for. It seems like with addons you can only call the main addon file and that would perform the routing necessary to those other commands.

Where as doing something like

```
fin kanopi/test
```

Having it as a command is more beneficial and easier for us to maintain if someone wants to add to the toolbox they don't run the risk of possibly messing up the router, it's also easier to maintain with having the commands be their own separate.

achekulaev:
@sean-e-dietrich gotcha. I think I see compromise for this PR. If you could change you code to only list executables in sub-folders, I would merge the PR.

sean-e-dietrich:
@achekulaev I needed to open a new PR for some odd reason but here is the new PR: https://github.com/docksal/docksal/pull/526

achekulaev:
@sean-e-dietrich thanks, merged that one, closing this one.

PR #483, opened 2018-03-23
run-cli (rc) should use 2.0 cli image.

ey-:
Trying out the run-cli command I figured out that it does not work with the new 2.0 images. I wanted to use it for a fast project start for a blt project like:
fin run-cli composer create-project --no-interaction acquia/blt-project my-project

But failed because that project has php 7.1 as requirement, which is not available for the 1.x images.

I tried to refactor a little bit to reduce the recurring if conditions but couldn't figure out how to optimize that and open for suggestions :)

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -3902,28 +3902,37 @@ run_cli ()
 	[[ $1 == "-T" ]] && \
 		local no_tty=true && shift
 	# Set default image
-	# Have to stick with docksal/1.3-php7 for now.
-	# TODO: Fix compatibility with docksal/cli:2.0 images.
-	local IMAGE="${IMAGE:-docksal/cli:1.3-php7}"
+	local IMAGE="${IMAGE:-docksal/cli:2.0-php7.1}"
 	local cmd="$@"
 
 	# Assume bash if no command was given
 	[[ "$cmd" == "" ]] && cmd="bash"
 
-	# Source $HOME/.docksalrc in cli.
-	# Commands in this file will be sourced for both interactive and non-interactive sessions.
-	DOCKSALRC='source $HOME/.docksalrc >/dev/null 2>&1'
-
 	# Debug mode off by default
 	DEBUG=${DEBUG:-0}
 
-	if is_tty && [[ "$no_tty" != true ]]; then
-		# interactive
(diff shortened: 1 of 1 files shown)


lmakarov:
Hey, @ey- !  I have not spent much time trying to figure out the compatibility between `run-cli` and v2 images. Thanks for looking into this.

I suggest we just drop support for using 1.x image with `run-cli` altogether to keep things simple.
If someone tries to use a 1.x image here, we throw and error and explain that that is not supported.

ey-:
Here you are @lmakarov. Support for v1 images are dropped & displaying an error instead.

bin/fin
```diff
@@ -3902,28 +3902,27 @@ run_cli ()
...
 
-	# Source $HOME/.docksalrc in cli.
-	# Commands in this file will be sourced for both interactive and non-interactive sessions.
-	DOCKSALRC='source $HOME/.docksalrc >/dev/null 2>&1'
-
 	# Debug mode off by default
 	DEBUG=${DEBUG:-0}
 
+	# Fail for images from v1, these are executing the commands directly and are not supported anymore.
+	if [[ "$IMAGE" == docksal/cli:1* ]]; then
```
achekulaev: There should be a period after 1
```
if [[ "$IMAGE" == docksal/cli:1.* ]]; then
```

bin/fin
```diff
@@ -3902,28 +3902,27 @@ run_cli ()
...
-	local IMAGE="${IMAGE:-docksal/cli:1.3-php7}"
+	local IMAGE="${IMAGE:-docksal/cli:2.0-php7.1}"
 	local cmd="$@"
 
 	# Assume bash if no command was given
 	[[ "$cmd" == "" ]] && cmd="bash"
 
-	# Source $HOME/.docksalrc in cli.
-	# Commands in this file will be sourced for both interactive and non-interactive sessions.
-	DOCKSALRC='source $HOME/.docksalrc >/dev/null 2>&1'
```
achekulaev: @lmakarov don't we still require this file for nvm/npm?

ey-:
Hi @achekulaev ,
The missing dot is added.

The docksalrc file is added for the new cli images in the startup script itself. see https://github.com/docksal/service-cli/blob/master/7.1/startup.sh#L57

PR #496, opened 2018-04-05
Allow `fin alias` creation command to also update aliases

frederickjh:
Currently if one has an alias and they want to update it it must be first removed then added again. This pull request makes it possible to update a existing alias by using the normal fin alias creation command. A fin alias update command could have been created but for the limited number of times it would be used it make more sense to have this as part of the fin alias creation command.

I added the "-f" option to `ln` when creating an new fin alias link. This allow one to use the same command to update aliases vs. the current workflow of first removing the alias then adding it back in with the new path.

### Before this pull request
```bash
fin alias remove projectname
fin alias ~/docker/projectname projectname
```
### After this pull request
```bash
fin alias ~/docker/projectname projectname
```

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1511,10 +1511,10 @@ show_help_alias ()
 	echo
 	echo-green "Usage:"
 	printh "fin alias [list]" "Show aliases list"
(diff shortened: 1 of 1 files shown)


achekulaev:
Thanks!

PR #508, opened 2018-04-25
Fix db import for paths with spaces in their name

GMTA:
There were quotes missing around the input argument for ${pipe_cmd} causing `db import` to fail if the provided path argument contains spaces.


achekulaev:
Thanks @GMTA

PR #511, opened 2018-05-01
Allow dash and underscore in docker compose project name.

CyberJack:
The release of docker compose version 1.21.1 added support for dashes and underscores in the project name.
When updating the `COMPOSE_PROJECT_NAME_SAFE` variable differs from the name docker-compose uses.

See commit : https://github.com/docker/compose/commit/5fe3aff1c310d6b8684404b19a8246736cfe8f1a#diff-a04be36073678de95917341f90c2cf58

This change updates the name normalization to match the 1.21.1 version of docker-compose.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -53,8 +53,8 @@ is_mac ()
 
 DOCKSAL_VERSION="${DOCKSAL_VERSION:-master}"
 
-REQUIREMENTS_DOCKER='17.09.0-ce'
-REQUIREMENTS_DOCKER_COMPOSE='1.19.0'
+REQUIREMENTS_DOCKER='18.03.1-ce'
+REQUIREMENTS_DOCKER_COMPOSE='1.21.1'
 REQUIREMENTS_DOCKER_MACHINE='0.14.0'
 REQUIREMENTS_VBOX='5.1.28'  # Remember to update the download URLs below
 REQUIREMENTS_WINPTY='0.4.3'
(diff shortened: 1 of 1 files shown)


achekulaev:
Hi Peter

I'm not sure it's a good idea because `VIRTUAL_HOST` usually equals to `COMPOSE_PROJECT_NAME_SAFE`. And I find that a useful default convention. 
However `VIRTUAL_HOST` can not contain underscores.
Please see https://blog.docksal.io/virtual-host-value-limitations-4f105725dbf8

So unless there is a very solid argumentation in favor or allowing underscore in compose name only and having virtual host differ from it in this case (because we would have to cut underscore out) I can not accept this PR.

I'm putting it into pending mode for discussion.

achekulaev:
@CyberJack yes, thanks for pointing all that out. 

If you could add removing underscore when forming `VIRTUAL_HOST` to the Pull Request, then I'd merge it right away. 

Virtual Host is set in couple fo places. 

In `load_configuration`

```bash
export VIRTUAL_HOST=...
```

and in `project_create` (this one is actually a bug :) ) where we need to sanitize `project_name`

```bash
echo "VIRTUAL_HOST=\"${project_name}...
```

CyberJack:
I added added the separate normalization of project- and virtual host name.
Also the `Docker` and `docker-compose` version have been updated to match the version in which Docker changed the normalization of project names.

achekulaev:
@lmakarov although we were not looking to update Docker version because of https://github.com/moby/moby/issues/35933 I'm inclined to accept this change in scope of this PR for 2 reasons

1. 18.03.1 seems to have some [potential remedy](https://github.com/moby/moby/issues/35933#issuecomment-386811596) for that issue
2. 17.09 seems pretty far behind now

lmakarov:
Merged. @CyberJack thank you for this contribution!

PR #519, opened 2018-05-04
Adding Environment Variables to run-cli command

sean-e-dietrich:
Added in ability for run-cli command to accept docker -e variables.

The purpose for this is that by default variables like `SECRET_TERMINUS_TOKEN` are not passed to the container nor are there currently an easy way to pass them to the `run-cli` container. Additionally, if there are custom images that are use it may be necessary to pass on those environmental variables.

Example use case now:

```
SECRET_TERMINUS_TOKEN=xxx;
fin run-cli -T -e SECRET_TERMINUS_TOKEN=$SECRET_TERMINUS_TOKEN 'terminus sites --fields="id,name,framework" --format=csv';
```

This will allow for us to automatically login to terminus (which when the variable is presented automatically does) and then run any necessary commands through the run-cli command.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -3909,25 +3909,17 @@ run_cli ()
 		-T) local no_tty=true; shift; ;;
 		# Throw all environment variables into an array so that they can be spit into the docker run command.
 		-e)
-		local vars=($(echo $2 | tr "=" " "))
-		local key=${vars[0]}
-		vars=("${vars[@]:1}")
-		local value=${vars[@]}
+		IFS='=' read -r key value <<< "${2}"
 
-		# Test if $key is already a defined variable if not add 
-		# to $env_str which will be passed to the docker run command.
-		if [[ -z ${!key+x} ]]; then
-			if [[ -z "${value}" ]]; then
-				env_str="${env_str} -e ${key}"
-			else
-				env_str="${env_str} -e ${key}='${value}'"
-			fi
-		else
-			if [[ ! -z "${value}" ]]; then
-				eval ${key}=`echo -ne \""${value}"\"`
-			fi
-			env_str="${env_str} -e ${key}"
+		# If value is empty set the value to whatever the variable name in $key is
+		# If key does not exist then sets it to empty variable
+		if [[ -z $value ]]; then
+			value="${!key}"
 		fi
+		# Set the $key back to whatever $value is
+		eval ${key}=`echo -ne \""${value}"\"`
+		# Add only key to $env_str to be used at runtime.
+		env_str="${env_str}-e ${key} "
 		shift 2
 		;;
 		*) break;;
--- a/bin/fin
+++ b/bin/fin
@@ -3898,26 +3898,43 @@ _exec ()
 # @param $* command with its params to run.
 run_cli ()
 {
-	local env_vars=()
 	local env_str=""
+
+	local HOST_UID=$(id -u)
+	local HOST_GID=$(id -g)
+
+	[[ -f "$CONFIG_ENV" ]] && set -a; source "$CONFIG_ENV";
 
 	while [ $# -gt 0 ]
 	do
-		case $1 in
-			# Allow disabling TTY mode.
-			# Useful for non-interactive commands when output is saved into a variable for further comparison.
-			# In a TTY mode the output may contain unexpected control symbols/etc.
-			-T) local no_tty=true; shift; ;;
-			# Throw all environment variables into an array so that they can be spit into the docker run command.
-			-e) env_vars+=("${2}"); shift 2; ;;
-			*) break;;
-		esac
-	done
+	    case $1 in
+		# Allow disabling TTY mode.
+		# Useful for non-interactive commands when output is saved into a variable for further comparison.
+		# In a TTY mode the output may contain unexpected control symbols/etc.
+		-T) local no_tty=true; shift; ;;
+		# Throw all environment variables into an array so that they can be spit into the docker run command.
+		-e)
+		local vars=($(echo $2 | tr "=" " "))
+		local key=${vars[0]}
+		vars=("${vars[@]:1}")
+		local value=${vars[@]}
 
-	# If env_vars array has anything in it let's put it into a string.
-	if [ ${#env_vars[@]} -gt 0 ]; then
-		local env_str=$(printf " -e %s " "${env_vars[@]}")
-	fi
+		# Test if $key is already a defined variable if not add 
+		# to $env_str which will be passed to the docker run command.
+		if [[ -z ${!key+x} ]]; then
+			if [[ -z "${value}" ]]; then
+				env_str="${env_str} -e ${key}"
+			else
+				env_str="${env_str} -e ${key}='${value}'"
+			fi
+		else
+			eval ${key}=`echo -ne \""${value}"\"`
+		fi
+		shift 2
+		;;
+		*) break;;
+	    esac
+	done
 
 	# Set default image
 	local IMAGE="${IMAGE:-docksal/cli:2.1-php7.1}"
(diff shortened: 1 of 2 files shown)


lmakarov:
@sean-e-dietrich that's for bringing up this use case and PR!

I think, we should align the variables with the `cli` service configuration in `services.yml`:

```
    environment:
      - HOST_UID
      - HOST_GID
      - DOCROOT
      - VIRTUAL_HOST
      - XDEBUG_ENABLED=${XDEBUG_ENABLED:-0}
      - XDEBUG_CONFIG=remote_connect_back=0 remote_host=${DOCKSAL_HOST_IP}  # Point xdebug to the host IP
      - SECRET_SSH_PRIVATE_KEY
      - SECRET_ACAPI_EMAIL
      - SECRET_ACAPI_KEY
      - SECRET_TERMINUS_TOKEN
```

With the exception of xdebug variables, all others should be passed in `run-cli`.

Allowing passing custom variables in `run-cli` would not hurt either. We'll need fin help and docs updated as well.

sean-e-dietrich:
@lmakarov totally agree! I made the modification and included those variables you mentioned but also included the ability to include items from the `$HOME/.docksal/docksal.env`. 

I included those variables but then included the ability to override them over the command line. I also included some BAT tests.

Example:

If `SECRET_TERMINUS_TOKEN` is in the `$HOME/.docksal/docksal.env` file but a person wanted to override in the command.

```
fin run-cli -T -e SECRET_TERMINUS_TOKEN=$SECRET_TERMINUS_TOKEN 'terminus sites --fields="id,name,framework" --format=csv';
```

lmakarov:
@sean-e-dietrich I was going to push some updates, but also did a rebase and looks like I killed a change you pushed, as we were working on it at the same time. I'm done with the updates now and waiting for tests to complete.

bin/fin
```diff
@@ -3900,11 +3900,6 @@ run_cli ()
 {
 	local env_str=""
 
-	local HOST_UID=$(id -u)
-	local HOST_GID=$(id -g)
-
-	[[ -f "$CONFIG_ENV" ]] && set -a; source "$CONFIG_ENV";
```
sean-e-dietrich: @lmakarov I'm curious why we don't want to have this in here? This would allow for variables from the home `docksal.env` file. Such as storing a Terminus Token or Acquia Credentials.

sean-e-dietrich: By eliminating this line then this doesn't test doesn't work proplery

```
 	echo "TEST=1234" >> $HOME/.docksal/docksal.env
 	run fin rc -T -e TEST 'echo \$TEST'
 	[[ "$output" == "1234" ]]
 	unset output
```

sean-e-dietrich: I apologize I misread the chain of events and found in the `RUNTIME STARTS HERE` is where the file is sourced.

lmakarov: The global `docksal.env` is already sourced at the very beginning of runtime in `fin` (https://github.com/docksal/docksal/blob/develop/bin/fin#L5000), so `SECRET_` variables will be passed without issues.

`HOST_UID` and `HOST_UID` used to be in `load_configuration`. I moved them to the runtime section as well.

Tests look good.

achekulaev:
Hi. I'm still looking into it, because I see there is a custom variables parser written in this function to parse variables from the command line. 
However there is already a variable parser existing in fin. It's `parse_params` function. 

```
eval $(parse_params "$@")
```

I like the contribution and I wanted to just change the PR myself to use the `parse_params` parser, but I have not got to it just yet.

@sean-e-dietrich if you could see how `parse_params` is used in the code and change PR to use it instead of the custom one you have created that would be awesome!

sean-e-dietrich:
@achekulaev specifically with the `parse_params` is 1 it takes the values of the params and sets them to their as their own values.

So specifically if I have the following:

```
fin run-cli -T -e var1=123 -e var2=2345 "echo \$var1"
```

It will make `var1` it's own variable, and `var2` and technically I could go through each of the variables  and use the `parse_params` function.

But if I run the following:

```
 fin run-cli -e TEST=123 -e TEST1=2345 -e SECRET_TERMINUS_TOKEN "echo \$TEST"
```

It doesn't pick `SECRET_TERMINUS_TOKEN` as a variable but adds it to the `ARGV` array therefore when I need to pass to `docker run` it wouldn't get pass properly.

The goal is at the time of running `fin run-cli` you could choose to have variables in the global `docksal.env` or provide them at run time, or override the ones in `docksal.env` at runtime.

Examples:

```
	# check to make sure custom variables are passed into container
	run fin rc -T -e TEST_VAR="TEST VARIABLES" 'echo \$TEST_VAR'
	[[ "$output" == "TEST VARIABLES" ]]
	unset output

	# check to make sure a global default variable (from $HOME/.docksal/docksal.env) is passed automatically.
	# These are SECRET_ and some other variables passed by default.
	echo "SECRET_SSH_PRIVATE_KEY=xyz" >> $HOME/.docksal/docksal.env
	run fin rc -T 'echo \$SECRET_SSH_PRIVATE_KEY'
	[[ "$output" == "xyz" ]]
	unset output

	# Check to make sure a global default variable can be overridden
	run fin rc -T -e SECRET_SSH_PRIVATE_KEY="abc" 'echo \$SECRET_SSH_PRIVATE_KEY'
	[[ "$output" == "abc" ]]
	unset output

	# check to make sure a global (non-default) variable can be passed if included in command
	echo "TEST=1234" >> $HOME/.docksal/docksal.env
	run fin rc -T -e TEST 'echo \$TEST'
	[[ "$output" == "1234" ]]
	unset output
```

If you would like `parse_params` to be used the it needs to be refactored to allow for non set variables.

Example:

```
fin run-cli -e TEST=123 -e TEST1=2345 -e TEST2 "echo \$TEST"

DEBUG
---------
local ARGV=();
-e
local TEST="123";
-e
local TEST1="2345";
-e
ARGV+=('TEST2');
ARGV+=('echo $TEST');
```

If I use this function now then it looks like `TEST2` would never get added as a variable to the `docker run` command.

I'm also unsure how you would see this fitting in @achekulaev. I also am unsure how you would grab the key of that variable and pass it onto the command


```
        # String of data to be passed 
	local env_str=""

        # Loop through all  variables
	while [ $# -gt 0 ]
	do
             # check to see if a flag is being used
	    case $1 in
		# Allow disabling TTY mode.
		# Useful for non-interactive commands when output is saved into a variable for further comparison.
		# In a TTY mode the output may contain unexpected control symbols/etc.
		-T) local no_tty=true; shift; ;;
		# Throw all environment variables into an array so that they can be spit into the docker run command.
		-e)
                 # Put break variable at the =
		local vars=($(echo $2 | tr "=" " "))
                 # Key is always first value
		local key=${vars[0]}
		vars=("${vars[@]:1}")
                 # Value is everything after
		local value=${vars[@]}

		# Test if $key is already a defined variable if not add
		# to $env_str which will be passed to the docker run command.
                 # if key has not already been defined in either global or run time
		if [[ -z ${!key+x} ]]; then
                         # If the value is empty
			if [[ -z "${value}" ]]; then
                                  # Add key to command
				env_str="${env_str} -e ${key}"
			else
                                  # Set key on command string
				env_str="${env_str} -e ${key}='${value}'"
			fi
		else
                         # If value of key is not empty
			if [[ ! -z "${value}" ]]; then
                                  # set key to value
				eval ${key}=`echo -ne \""${value}"\"`
			fi
                         # Add key to runtime
			env_str="${env_str} -e ${key}"
		fi
		shift 2
		;;
		*) break;;
	    esac
	done
```

achekulaev:
@sean-e-dietrich thanks for a detailed reply, as usual. Yes indeed `parse_params` is not suited to work with params, that have no value passed. 

So if we will need to leave the new parser, I would like to simplify some things. Despite the code is completely functional, it is somewhat hard to maintain for some operations look too complicated.

1. Could you simplify this:

```bash
local vars=($(echo $2 | tr "=" " "))
local key=${vars[0]}
vars=("${vars[@]:1}")
local value=${vars[@]}
```

to this

```bash
local key="$(echo $2 | cut -f1 -d=)"
local value="$(echo $2 | cut -f2 -d=)"
```

2. And this:

```bash
if [[ -z ${!key+x} ]]; then
```

to this

```bash
if [[ -n "$key" ]]; then
```

PR #523, opened 2018-05-08
Turn $VIRTUAL_HOST output in terminal into link for "fin up"

frederickjh:
Currently after running `fin up` at the end we see the contents of $VIRTUAL_HOST printed to the screen, but this is not a link adding **http://** before the variable output will make this a clickable link on GUI terminals that can then be easily opened.

I added **http://** before the $VIRTUAL_HOST output in terminal for `fin up`. This makes it a link that is clickable. Normally by using **Ctrl-click** or **right-click** to bring up a context sensitive menu.

#### Before
Not a link.

image
#### After
Clickable link that is underlined on mouse over and right-click give context sensitive menu with options **Open Link** and **Copy Link Address**.

image

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -2409,7 +2409,7 @@ up ()
 
 	_start_containers
 	unison_sync_wait
-	echo-green "Virtual Host: ${yellow}$VIRTUAL_HOST${NC}"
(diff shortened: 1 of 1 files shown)


achekulaev:
Thanks.

Issue #524, opened 2018-05-09
Document addons.


opdavies:
Add documentation page for addons.

achekulaev:
Done by @sean-e-dietrich in https://github.com/docksal/docksal/pull/528

PR #528, opened 2018-05-12
Addon docs

sean-e-dietrich:
Added documentation for Addons this should satisfy #524 which asks for documentation for the addons.

I placed the command in the same structure as the Custom Commands and within the `fin` command as I figured that is most likely the best place for that to go.


achekulaev:
This is awesome. Thanks Sean!

PR #533, opened 2018-05-16
Extending NGROK to use Variables

sean-e-dietrich:
Extending NGROK so that it utilizes the docker environment variables for authentication.

Environmental variables are outlined on the README within the [repo](https://github.com/wernight/docker-ngrok#environment-variables)

Particularly this is useful if someone has an NGROK account and wants to authenticate using that.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -4450,6 +4450,14 @@ ngrok_share ()
 		--net ${network} \
 		--link ${container_name} \
 		--name ${ngrok_container_name} \
+		-e NGROK_AUTH \
+		-e NGROK_SUBDOMAIN \
+		-e NGROK_HOSTNAME \
+		-e NGROK_USERNAME \
+		-e NGROK_PASSWORD \
+		-e NGROK_PROTOCOL \
+		-e NGROK_PORT \
+		-e NGROK_REGION \
 		wernight/ngrok \
 		ngrok http -host-header "${host:-$NGROK_VIRTUAL_HOST}" ${container_name}.${network}:80
 }
--- a/bin/fin
+++ b/bin/fin
@@ -4449,44 +4449,43 @@ ngrok_share ()
 	# and use -e variables on the docker container
 	# Modifying entrypoint.sh file to gain more variables.
 
-	ARGS="ngrok"
+	local ARGS="ngrok"
 	# Set the protocol.
-	if [ "$NGROK_PROTOCOL" = "TCP" ]; then
+	if [[ "$NGROK_PROTOCOL" = "TCP" ]]; then
 	  ARGS="$ARGS tcp"
 	else
 	  ARGS="$ARGS http"
 	  NGROK_PORT="${NGROK_PORT:-80}"
 	fi
+	
+	NGROK_PORT=$(echo ${NGROK_PORT} | sed 's|^tcp://||')
 
 	# Set the authorization token.
-	if [ -n "$NGROK_AUTH" ]; then
+	if [[ -n "$NGROK_AUTH" ]]; then
 	  ARGS="$ARGS -authtoken=$NGROK_AUTH "
 	fi
 
 	# Set the subdomain or hostname, depending on which is set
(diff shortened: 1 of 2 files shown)


bin/fin
```diff
@@ -4445,13 +4445,68 @@ ngrok_share ()
...
+	  ARGS="$ARGS -region=$NGROK_REGION "
+	fi
+
+	if [ -n "$NGROK_HEADER" ]; then
+	  ARGS="$ARGS -host-header=$NGROK_HEADER "
+	fi
+
+	if [ -n "$NGROK_USERNAME" ] && [ -n "$NGROK_PASSWORD" ] && [ -n "$NGROK_AUTH" ]; then
+	  ARGS="$ARGS -auth=\"$NGROK_USERNAME:$NGROK_PASSWORD\" "
+	elif [ -n "$NGROK_USERNAME" ] || [ -n "$NGROK_PASSWORD" ]; then
```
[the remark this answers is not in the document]
achekulaev: As far as I understand the message, all 3 should be set for some ngrok functionality that requires auth to work properly, and if only 2/3 or 1/3 is set, then that ngrok f-ty will not function properly. 

If this my understanding is correct, then current if-case with the change that I proposed above should be fine.

```bash
if [[ -n "$NGROK_USERNAME" ]] && [[ -n "$NGROK_PASSWORD" ]] && [[ -n "$NGROK_AUTH" ]]; then
  ARGS="$ARGS -auth=\"$NGROK_USERNAME:$NGROK_PASSWORD\" "
elif [[ -n "$NGROK_USERNAME" ]] || [[ -n "$NGROK_PASSWORD" ]] || [[ -n "$NGROK_AUTH" ]]; then
  echo "..."
  return 1
fi
```

If all 3 are empty then it is fine and no message needed.

bin/fin
```diff
@@ -4445,13 +4445,68 @@ ngrok_share ()
...
+
+	# Set the subdomain or hostname, depending on which is set
+	if [ -n "$NGROK_HOSTNAME" ] && [ -n "$NGROK_AUTH" ]; then
+	  ARGS="$ARGS -hostname=$NGROK_HOSTNAME "
+	elif [ -n "$NGROK_SUBDOMAIN" ] && [ -n "$NGROK_AUTH" ]; then
+	  ARGS="$ARGS -subdomain=$NGROK_SUBDOMAIN "
+	elif [ -n "$NGROK_HOSTNAME" ] || [ -n "$NGROK_SUBDOMAIN" ]; then
+	  if [ -z "$NGROK_AUTH" ]; then
+	    echo "You must specify an authentication token after registering at https://ngrok.com to use custom domains."
+	    exit 1
```
achekulaev: Please use `return 1`

bin/fin
```diff
@@ -4445,13 +4445,68 @@ ngrok_share ()
...
+	  ARGS="$ARGS -host-header=$NGROK_HEADER "
+	fi
+
+	if [ -n "$NGROK_USERNAME" ] && [ -n "$NGROK_PASSWORD" ] && [ -n "$NGROK_AUTH" ]; then
+	  ARGS="$ARGS -auth=\"$NGROK_USERNAME:$NGROK_PASSWORD\" "
+	elif [ -n "$NGROK_USERNAME" ] || [ -n "$NGROK_PASSWORD" ]; then
+	  if [ -z "$NGROK_AUTH" ]; then
+	    echo-yellow "You must specify a username, password, and Ngrok authentication token to use the custom HTTP authentication."
+	    echo-yellow "Sign up for an authentication token at https://ngrok.com"
+	    exit 1
```
achekulaev: Please use `return 1`

sean-e-dietrich:
@achekulaev I reworked the script as recommended. Thanks for the feed back part of it was I just copied and pasted what was in the entry script for the docker image. But the changes you recommended were good thank you!

achekulaev:
@sean-e-dietrich thanks! Perfect!

Issue #405, opened 2017-11-29
Support BitBucket as addons source


achekulaev:
Currently addons can only be installed from GitHub repositories. This is a ticket to add BitBucket support. Succeeds #368 

GitHub `https://raw.githubusercontent.com/$USER/$REPO/$BRANCH/$PATH/$FILE` 
BitBucket `https://bitbucket.org/$USER/$REPO/raw/$BRANCH/$PATH/$FILE`

lmakarov:
See https://github.com/docksal/docksal/issues/368#issuecomment-348086947. I think should instead add support for arbitrary git repos (public or private).

shelane:
Agreed. We have an on prem Bitbucket server where we would host ours.

sean-e-dietrich:
@shelane we are currently looking at creating a way to include your own repositories in the PR https://github.com/docksal/docksal/pull/539 there have just been more pressing issues at the moment. But that is good to know this is a feature that others would like.

Issue #368, opened 2017-09-27
Allow for private addon repo via ENV var


vml-bsummers:
if URL_REPO_ADDONS could be configured in docksal.env then I would be able to host private addons available within my company.

achekulaev:
Functionality added to the `develop` branch in the commit above.

The source repo is controlled not via the environment variable though, because I would like to keep experience of installing addons from the Docksal repo streamlined and with environment variable in could easily be disrupted.

However now you can install addons from a different repo by providing repo path in the addon install command. 

Example:
```
fin addon install achekulaev/docksal-addons/uli
```

Of course it requires the source repo (in this case https://github.com/achekulaev/docksal-addons/) to follow the same structure the https://github.com/docksal/addons repo is following. Addons can only be installed from the `master` branch at the moment.

vml-bsummers:
Does this require the additional repo to be on github? 

What about just a secondary repo to check if not found in the primary add-on repo?

achekulaev:
Yes it only supports Github at the moment. Other providers have very different URL structures. Which one you are interested in?

> What about just a secondary repo to check if not found in the primary add-on repo?

I feel a security issue with such approach.

vml-bsummers:
We use hosted bitbucket. But this is all just how I would want to use such a feature. I'm not really considering other uses or abuses

achekulaev:
I guess hosted bitbucket is going to be a popular option. I would create a separate ticket for that. It will require additional work. 
Currently addons are installed by pulling the files directly. fin does not clone the whole repo just to install a single addon. That's why it is dependent on the URLs structure of the repo cloud provider.
So for GitHub it's one:
`https://raw.githubusercontent.com/$USER/$REPO/$BRANCH/$PATH/$FILE`
While for BitBucket it is different 
https://bitbucket.org/$USER/$REPO/raw/$COMMIT_HASH/$PATH/$FILE
and I guess hosted BitBucket it is yet different from that and I don't even have the latter at the moment to test.

lmakarov:
@achekulaev 8887096 expands the scope of where an addon can be pulled from on Github, but it still has to sit a **public** repo.

I believe this issue is about supporting **arbitrary** and/or **private** git repos.

That can be achieved if either: 

- every addon sits in it's own repo (like custom Drupal modules or any other library over there). I'm not a fan of this idea.
- the while "addons" repo is pulled with git every time to get only a specific addon from it. While it sounds inefficient, I don't this this would be such a bad idea.

[...]

PR #535, opened 2018-05-17
Addon Repositories Extension.

sean-e-dietrich:
Allow for multiple repositories to be used when searching for addons. Updated documentation to reflect changes in PR.

The change allows for multiple repositories to be used. Additionally, it allows for the ability to use private repos.

I believe this solves #368 and solves #405 

The main purpose for this is if I had a private repository for my team then I could install from that. Additionally, I could check against multiple repositories, sort of like a package manager.

The URL structure would be contained in the `.docksal/docksal.env` file and look like this:

```
DOCKSAL_ADDON_REPO_TEAM_REPO="URL='https://raw.githubusercontent.com/sean-e-dietrich/addons';BRANCH='develop'"
```
Additional addon repos could be added by prefixing their variables with `DOCKSAL_ADDON_REPO_` which would then search before hand to make sure the addon existed before continuing.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1728,11 +1728,13 @@ show_help_addon ()
 	echo-green "Commands:"
 	printh "install <name>" "Install addon"
 	printh "remove <name>" "Remove addon"
+	printh "add-repo REPO=<repo-url> BRANCH=<repo-branch> AUTH=<repo-auth>" "Add Addon Repository"
 
 	echo
 	echo-green "Examples:"
 	printh "fin addon install solr" "Install solr addon to the current project"
 	printh "fin addon remove solr" "Uninstall solr addon from the current project"
+	printh "fin addon add-repo REPO=\"https://raw.githubusercontent.com/test-user/docksal-toolkit\" BRANCH=develop AUTH=\"github_user:github_token\"" "Add Repository for test-user/docksal-toolkit using only the develop branch and using the specific credentials"
 }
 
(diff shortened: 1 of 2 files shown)


achekulaev:
Hi Sean

Didn't know you wanted to enhance that part, but since we are there already... 

It is long to explain reasons here, so I'll just jump to conclusions. The way I imagined addon installation initially and the way people really need it played out to be different.

That's why what I really want to do is to rewrite addons installation completely and instead of downloading separate files perform a shallow git clone (`--depth=1`) of the target repo into `/tmp` directory and then copy over the desired addon into the project. The default repo would stay the `docksal/addons` one still. And the default in-repo path would still be `/addons` unless you have proposition for improvement since you real-project experience with that use-case.

This would help to support all possible Git providers and authentication methods, since we would be using `git clone`.

ENV vars names would respectively be:

```
DOCKSAL_ADDONS_REPO
DOCKSAL_ADDONS_REPO_BRANCH
```

This is a pretty major refactoring that will include getting rid of vars like `URL_ADDONS_HOSTING` and `URL_ADDONS_REPO` and changing logics of `addon_install` completely and maybe affecting some other `addon_*` funcitons.

If you're open to doing that, I could invite you into the Docksal Slack so we could discuss there in more details or maybe even have a call.

cc @lmakarov

sean-e-dietrich:
Closing in Favor of #539

PR #538, opened 2018-05-20
Included DOCROOT on config generate

sean-e-dietrich:
Exporting the DOCROOT variable on fin config generate as it may not be the standard docroot folder.

When running `DOCROOT=web fin config generate` within a project it doesn't include the DOCROOT within `docksal.yml` when the DOCROOT folder is not named `docroot`.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -4854,13 +4854,13 @@ config_generate ()
 	mkdir -p ".docksal"
 	
 	# Set Stack at Config Run Time
-	local STACK=${STACK:-default}
+	local STACK=${stack:-default}
 	
 	# Create docksal.env if it does not exist
 	touch ".docksal/docksal.env" && \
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -1661,7 +1661,7 @@ show_help_config ()
 	printh "show [options]" "Display configuration for the current project"
 	printh "    --show-secrets" "Do not truncate value of SECRET_* environment vars"
 	printh "env" "Display only environment variables section"
-	printh "generate" "Generate empty Docksal configuration for the project"
+	printh "generate [STACK=VAL] [DOCROOT=VAL]" "Generate empty Docksal configuration for the project"
```
achekulaev: Let's use the default notation of `--stack=...` `--docroot=...` here

PR #539, opened 2018-05-21
Multiple Repositories for Addons

sean-e-dietrich:
Started development for multiple repositories that can be used for addons.
Additionally, added ability to run updates between addons. Rather then
running the risk of the addons being out of sync

This is a rewrite of the PR #535 and I believe this solves #368 and solves #405

This takes information that was written through back channels and allows for GIT repos to be used when checking out addons instead of downloading raw files with only support for github.

Additionally, support for multiple repositories is added. The repositories must be prefixed with `DOCKSAL_ADDONS_REPO_` in order to have them searched. The `DOCKSAL_ADDONS_REPO` which is the same repo at https://github.com/docksal/addons is searched first but can be overridden within the `.docksal/docksal.env` file.

Syntax:
```
DOCKSAL_ADDONS_REPO_[NAME]="[Repo URL]|[Branch]"
```

Because a repository can have multiple branches there is an additional options for the variable to designate which branch is selected.

`[NAME]` Should be replaced with a unique machine name for that repo
`[Repo URL]` Should be the GIT repo url that will be used to clone the repo.
`[BRANCH]` Is optional but designates what branch of the repo to use, if not included `master` branch will be used.

**NOTE** A pipe `(|)` is used to separate the `[REPO URL]` and `[BRANCH]`. If `[BRANCH]` is omitted then the pipe is not necessary

Examples:
```
 # Checks master branch
DOCKSAL_ADDONS_REPO_TOOLKIT="https://github.com/kanopi/docksal-toolkit.git"
# Checks sample branch
DOCKSAL_ADDONS_REPO_TOOLKIT_SAMPLE="https://github.com/kanopi/docksal-toolkit.git|sample" 
```

This also adds in a `fin addon update` command which will allow for the addon to run a similar process as the `fin addon install`. The update process will look for `pre-update` and `post-update` hooks as well. 

The purpose for this is that if an addon has a single repo it can be tracked and to a certain extent versioned appropriately, but more so then that, if there is a need for developer to use their own personal addons in projects, but they also have a repo which is shared amongst a team then they can use both of the repositories.


achekulaev:
@sean-e-dietrich can you please rebase it on the develop branch that now has your commit to fix tests for external PRs?

sean-e-dietrich:
@achekulaev I already merged the changes in and am rerunning the tests.

bin/fin
```diff
@@ -167,6 +167,7 @@ URL_ADDONS_HOSTING="https://raw.githubusercontent.com"
 URL_ADDONS_REPO="$URL_ADDONS_HOSTING/docksal/addons"
 # Remove URL_REPO_ADDONS in Nov 2018
 URL_REPO_ADDONS="https://raw.githubusercontent.com/docksal/addons"
```
achekulaev: This variable is unused now

bin/fin
```diff
@@ -1728,11 +1729,13 @@ show_help_addon ()
 	echo
 	echo-green "Commands:"
 	printh "install <name>" "Install addon"
+    printh "update <name>" "Update addon"
```
achekulaev: Should use tab for indentation

bin/fin
```diff
@@ -5081,28 +5084,99 @@ addon () {
...
+
+# Search all of the provided repositories for the requested plugin.
+_addon_search_repo () {
+	eval 'repos=(${!'"DOCKSAL_ADDONS_REPO"'@})'
+
+	local addon=$1
+	local repo_url=$2
+	local repo_branch=$3
+	local repo_uuid=$4
+
```
achekulaev: This loop may take some time to execute. Let's warn user before it starts (only if repo count is > 1)

```
echo-green "Searching for ${addon} in predefined repositories..."
```

bin/fin
```diff
@@ -5081,28 +5084,99 @@ addon () {
...
+	for key in "${repos[@]}"
+	do
+		local config=${!key}
+		# Append master just incase there is no branch defined
+		config="${config}|master"
+		local REPO_URL=$(echo $config | cut -d'|' -f1)
+		local REPO_BRANCH=$(echo $config | cut -d'|' -f2)
+        local uuid=$(uuidgen)
```
achekulaev: Please use tabs everywhere for indentation

bin/fin
sean-e-dietrich: I can remove it, I was leaving it there because I know you said remove in Nov '18

bin/fin
```diff
@@ -5081,28 +5084,99 @@ addon () {
...
+
+        # Check to make sure the provided Repo URL is valid and the Branch Exists
+        local valid=$(git ls-remote -h $REPO_URL --refs $REPO_BRANCH | wc -l)
```
achekulaev: 1. git's stderr should be redirected to dev/null
2. since it's non-fatal I would avoid echo-error
3. it's better to use `grep -q $REPO_BRANCH` to avoid counting some random output as a valid output

```bash
# Check that provided repo URL is valid and the branch exists
git ls-remote -h "$REPO_URL" --refs "$REPO_BRANCH" 2>/dev/null | grep -q "$REPO_BRANCH"
if [[ "$?" != "0" ]]; then
  echo "$key defines an invalid repository/branch reference. Skipping..." 
  continue
else
...
```

bin/fin
achekulaev: Please don't use uuidgen. I'd like to avoid relying on additional binaries when possible. Use `$RANDOM` bash variable. Search for `/tmp/etc.exports.$RANDOM` in fin to see usage example

bin/fin
```diff
@@ -5081,28 +5084,99 @@ addon () {
...
+    local repo_branch=${4}
+    local tmp_random="/tmp/${uuid}"
+    mkdir $tmp_random && cd $tmp_random
+
+    # Initialize empty repo; suppress output
+    git init > /dev/null
+    git remote add origin $repo_url > /dev/null
+    git fetch -q origin $repo_branch > /dev/null
+    git checkout origin/${repo_branch} ${addon_name} 2> /dev/null
+    git checkout origin/${repo_branch} ${addon_name}.pre-install 2> /dev/null
```
achekulaev: Not every addon has pre-install hook. We should only check the main file

bin/fin
```diff
@@ -5081,28 +5084,99 @@ addon () {
...
+    local repo_url=${3}
+    local repo_branch=${4}
+    local tmp_random="/tmp/${uuid}"
+    mkdir $tmp_random && cd $tmp_random
+
+    # Initialize empty repo; suppress output
+    git init > /dev/null
+    git remote add origin $repo_url > /dev/null
+    git fetch -q origin $repo_branch > /dev/null
+    git checkout origin/${repo_branch} ${addon_name} 2> /dev/null
```
achekulaev: I think this command should actually be the last. If file did check out then everything is in order. If it did not then it is an error.

bin/fin
```diff
@@ -5127,85 +5206,125 @@ addon_install ()
...
-			if_failed_error "Download has failed" "Check log above for messages"
-		fi
+		echo-red "Addon already exists in $addon_path. Run ${NC}fin addon update ${addon_name}"
+		exit 1
+	else
+		mkdir -p "$addon_path"
+		if_failed_error "Could not create $addon_path" "Check file permissions and try again"
+	fi
+
+	local tmp_addon="/tmp/${ADDON_UUID}"
```
achekulaev: Let refactor var name to `addons_repo_copy`

bin/fin
```diff
@@ -5127,85 +5206,125 @@ addon_install ()
...
-				for f in ${hooks_file_list}; do
-					echo "  $f"
-					# Exit subshell with error if download failed
-					curl -fsSL "$URL_ADDON/$f?$RANDOM" -o "$f" || exit 1
-				done
-			)
-			if_failed_error "Download has failed" "Check log above for messages"
-		fi
+		echo-red "Addon already exists in $addon_path. Run ${NC}fin addon update ${addon_name}"
+		exit 1
```
achekulaev: return 1

bin/fin
```diff
@@ -5127,37 +5203,31 @@ addon_install ()
...
+
+	local tmp_addon="/tmp/${ADDON_UUID}"
+
+	# If repo contains only 1 addon let's download the full repo.
+	if [[ -f $tmp_addon/$addon_name ]] && [[ -x $tmp_addon/$addon_name ]]; then
+		cd $tmp_addon/
+		git pull --depth=1 origin/${URL_ADDONS_BRANCH}
+		# Remove Git Repo that was cloned from.
+		rm -rf $tmp_addon/.git 2>/dev/null
+		mv -R $tmp_addon $addon_path
```
achekulaev: `mv -R` is an anti-pattern because of inability to properly handle errors (what failed, copy or delete?) and because it will leave leftovers if copying errors happen.
`cp -R` followed by `rm -rf` should be used instead.

bin/fin
```diff
@@ -5118,6 +5189,11 @@ addon_install ()
 	[[ "$addon_name" =~ \\ ]] && echo-error "No slashes in addon name" && exit 1
 	([[ "$addon_name" == ".." ]] || [[ "$addon_name" == "." ]]) && echo-error "Invalid addon name" && exit 1
 
+    local URL_ADDONS_REPO
+	local URL_ADDONS_BRANCH
+	local ADDON_UUID
+	_addon_search_repo $addon_name URL_ADDONS_REPO URL_ADDONS_BRANCH ADDON_UUID
```
achekulaev: This block should be moved down, below the check for addon existence

bin/fin
```diff
@@ -5127,37 +5203,31 @@ addon_install ()
...
+	if [[ -f $tmp_addon/$addon_name ]] && [[ -x $tmp_addon/$addon_name ]]; then
+		cd $tmp_addon/
+		git pull --depth=1 origin/${URL_ADDONS_BRANCH}
+		# Remove Git Repo that was cloned from.
+		rm -rf $tmp_addon/.git 2>/dev/null
+		mv -R $tmp_addon $addon_path
+	# If repo contains multiple addon let's only download the repo we need.
+	elif [[ -d $tmp_addon/$addon_name ]] && [[ -f $tmp_addon/$addon_name/$addon_name ]] && [[ -x $tmp_addon/$addon_name/$addon_name ]]; then
+		# Remove Git Repo that was cloned from.
+		rm -rf $tmp_addon/.git 2>/dev/null
```
achekulaev: Both outputs should be redirected

```bash
rm -rf "$tmp_addon/.git" >/dev/null 2>&1
```

bin/fin
```diff
@@ -5167,47 +5237,93 @@ addon_install ()
...
 	addon_hook "$addon_path" "$addon_name" "post-install" "$addon_global"
 	[[ $? != 0 ]] && _with_errors="${yellow}with errors${NC}"
 	[[ "$2" == "global" ]] && _globally="globally"
 	echo-green "Addon ${NC}$addon_name${green} was installed ${_globally} ${_with_errors}"
 }
 
+# Update addon
+# $1 - addon name
+# $2 - optional flag to install globally
+addon_update () {
```
achekulaev: How is this function different from addon_install?

bin/fin
```diff
@@ -5083,28 +5084,98 @@ addon () {
...
+			echo "$key defines an invalid repository/branch reference. Skipping..."
+			continue
+		else
+			local exists=$(_addon_temp_install $uuid $addon $REPO_URL $REPO_BRANCH)
+			if [[ -n "${exists}" ]] && [[ $exists -ge 0 ]]; then
+				if _confirm "Addon Found: Install ${NC}${key}/${addon}" "--no-exit" -eq 0; then
+					eval "${repo_url}=${REPO_URL}"
+					eval "${repo_branch}=${REPO_BRANCH}"
+					eval "${repo_uuid}=${uuid}"
+					break;
```
achekulaev: `return` ?

bin/fin
```diff
@@ -5083,28 +5084,98 @@ addon () {
...
+		config="${config}|master"
+		local REPO_URL=$(echo $config | cut -d'|' -f1)
+		local REPO_BRANCH=$(echo $config | cut -d'|' -f2)
+		local uuid=$(uuidgen)
+
+		# Check to make sure the provided Repo URL is valid and the Branch Exists
+		git ls-remote -h "$REPO_URL" --refs "$REPO_BRANCH" 2>/dev/null | grep -q "$REPO_BRANCH"
+		if [[ "$?" != "0" ]]; then
+			echo "$key defines an invalid repository/branch reference. Skipping..."
+			continue
```
achekulaev: There should be no continue here or `rm -rf "/tmp/${uuid}"` will not clean up

bin/fin
```diff
@@ -5083,28 +5084,98 @@ addon () {
...
+_addon_search_repo () {
+	eval 'repos=(${!'"DOCKSAL_ADDONS_REPO"'@})'
+
+	local addon=$1
+	local repo_url=$2
+	local repo_branch=$3
+	local repo_uuid=$4
+
+	echo-green "Searching for ${addon} in predefined repositories..."
+	for key in "${repos[@]}"
```
```
for repo in "${repos[@]}"
```

bin/fin
```diff
@@ -5083,28 +5084,98 @@ addon () {
...
+	do
+		local config=${!key}
```
achekulaev: no need for this, it should be `for config in "${repos[@]}"`

bin/fin
sean-e-dietrich: @achekulaev `addon_update` only handles the updating piece. I believe the goal would be to eventually make it so that there were `pre-update` and `post-update` hooks for addons. Eventually also there could be the capability of versioning.

bin/fin
```diff
@@ -5083,28 +5084,98 @@ addon () {
...
+				fi
+			fi
+		fi
+
+		rm -rf "/tmp/${uuid}"
+		unset REPO_URL REPO_BRANCH
+	done
+
+	if [[ -z "${REPO_URL}" ]]; then
+		echo-red "Addon not found to install."
```
```bash
echo-error "Could not find addon ${addon_name} in any of the pre-defined repositories." \
  "Check repositories and branches spelling"
return 1
```

bin/fin
```diff
@@ -5127,37 +5198,36 @@ addon_install ()
...
+		return 1
+	else
+		mkdir -p "$addon_path"
+		if_failed_error "Could not create $addon_path" "Check file permissions and try again"
+	fi
+
+	local URL_ADDONS_REPO
+	local URL_ADDONS_BRANCH
+	local ADDON_UUID
+	_addon_search_repo $addon_name URL_ADDONS_REPO URL_ADDONS_BRANCH ADDON_UUID
```
achekulaev: After this you need to check for results 

```bash
_addon_search_repo ...
[[ "$?" != "0" ]] && return 1
```

bin/fin
sean-e-dietrich: @achekulaev 

`repos` is an array of variable names
When looping through those variables, `key` refers to the variable name that is being used.
`config` is referring to the value for the variable that is stored.

Example:

```bash
DOCKSAL_ADDONS_REPO="https://github.com/docksal/addons"
DOCKSAL_ADDONS_REPO_X="https://github.com/kanopi/docksal-toolkit.git|sample"
eval 'repos=(${!'"DOCKSAL_ADDONS_REPO"'@})'
# should produce something like the following array
# repos=(DOCKSAL_ADDONS_REPO DOCKSAL_ADDONS_REPO_X)
for key in "${repos[@]}"
do
	local config=${!key} #when looped will now grab the value for that variable
done
```

achekulaev: Right. Then let's refactor to 
`for repo in "${repos[@]}"`
key is just too generic.

bin/fin
sean-e-dietrich: @achekulaev I'm not removing the folder here with the though that if they have chosen to install the module we will have the data downloaded and no need to re-clone it. Which is why I am passing the UUID as a reference so we know the folder to use.

bin/fin
```diff
@@ -5083,28 +5084,98 @@ addon () {
...
+				if _confirm "Addon Found: Install ${NC}${key}/${addon}" "--no-exit" -eq 0; then
+					eval "${repo_url}=${REPO_URL}"
+					eval "${repo_branch}=${REPO_BRANCH}"
+					eval "${repo_uuid}=${uuid}"
+					break;
+				fi
+			fi
+		fi
+
+		rm -rf "/tmp/${uuid}"
```
```bash
rm -rf "/tmp/${uuid}" >/dev/null 2>&1
```

bin/fin
```diff
@@ -5167,47 +5238,94 @@ addon_install ()
 		echo-red "Pre-install hook has failed and aborted the installation."
 		# Delete downloaded files if pre-install hook fails
 		[[ -d "$addon_path" ]] && [[ "$addon_path" =~ ".docksal" ]] &&
-			rm -r "$addon_path"
```
achekulaev: that indentation was intentional

bin/fin
achekulaev: I get that in future there might be the case, but is there any real code difference with `addon_install` right now? Because in PR review they look identical to me. And if they are identical right now I would just call `addon_install` for `fin addon update`

sean-e-dietrich: the only difference would be the hooks. So seeing as the same code is ran for them both maybe refactoring the code and putting it in a function of it's own so it isn't duplicate coding.

achekulaev:
@sean-e-dietrich could you give me write access to your repo fork? I'd like to download and test first and I think I will have more fixes afterwards.

sean-e-dietrich:
@achekulaev I couldn't give you access to the Kanopi's org but I cloned the repo into a personal account and made it private. Also added you on the repo as a collaborator

achekulaev:
Thanks. I don't need access to Kanopi.org, the access you've given to https://github.com/sean-e-dietrich/docksal which you have raised a PR from is enough.

lmakarov:
Needs to be rebased into develop.

bin/fin
```diff
@@ -5626,34 +5627,108 @@ addon () {
...
+# $1 - uuid to use for foldername in /tmp
+# $2 - addon_name the folder/file to find
+# $3 - repo_url the repository url to check
+# $4 - repo_branch the branch to check against
+_addon_temp_install () {
+	local uuid=${1}
+	local addon_name=${2}
+	local repo_url=${3}
+	local repo_branch=${4}
+	local tmp_random="/tmp/${uuid}"
```
```suggestion
	local tmp_random="/tmp/tmp.$RANDOM$RANDOM"
```

bin/fin
```diff
@@ -5626,34 +5627,108 @@ addon () {
...
 	esac
 }
 
+# Check to see if addon exists within remote repo
+# $1 - uuid to use for foldername in /tmp
+# $2 - addon_name the folder/file to find
+# $3 - repo_url the repository url to check
+# $4 - repo_branch the branch to check against
+_addon_temp_install () {
+	local uuid=${1}
```
```suggestion

```

bin/fin
```diff
@@ -5626,34 +5627,108 @@ addon () {
...
+		;;
 		*)
 			echo-yellow "Unknown sub-command. See ${NC}fin addon help${yellow} for details."
-			;;
+		;;
 	esac
 }
 
+# Check to see if addon exists within remote repo
+# $1 - uuid to use for foldername in /tmp
```
```suggestion

```

bin/fin
```diff
@@ -5626,34 +5627,108 @@ addon () {
...
+# $2 - addon_name the folder/file to find
+# $3 - repo_url the repository url to check
+# $4 - repo_branch the branch to check against
```
```suggestion
# $1 - addon_name the folder/file to find
# $2 - repo_url the repository url to check
# $3 - repo_branch the branch to check against
```

bin/fin
```diff
@@ -5626,34 +5627,108 @@ addon () {
...
+		local REPO_URL=$(echo $config | cut -d'|' -f1)
+		local REPO_BRANCH=$(echo $config | cut -d'|' -f2)
+		local uuid=$(uuidgen)
+
+		# Check to make sure the provided Repo URL is valid and the Branch Exists
+		git ls-remote -h "$REPO_URL" --refs "$REPO_BRANCH" 2>/dev/null | grep -q "$REPO_BRANCH"
+		if [[ "$?" != "0" ]]; then
+			echo "$repo defines an invalid repository/branch reference. Skipping..."
+		else
+			local exists=$(_addon_temp_install $uuid $addon $REPO_URL $REPO_BRANCH)
```
```suggestion
			local exists=$(_addon_temp_install $addon $REPO_URL $REPO_BRANCH)
```

bin/fin
```diff
@@ -5626,34 +5627,108 @@ addon () {
...
 
+# Check to see if addon exists within remote repo
+# $1 - uuid to use for foldername in /tmp
+# $2 - addon_name the folder/file to find
+# $3 - repo_url the repository url to check
+# $4 - repo_branch the branch to check against
+_addon_temp_install () {
+	local uuid=${1}
+	local addon_name=${2}
+	local repo_url=${3}
```
```suggestion
	local repo_url=${2}
```

bin/fin
```diff
@@ -5626,34 +5627,108 @@ addon () {
...
+	local repo_branch=${4}
```
```suggestion
	local repo_branch=${3}
```

lmakarov:
This PR has been open for 2.5 years. @sean-e-dietrich do you still intend finalizing it? If not, please close.

Issue #517, opened 2018-05-03
Update web image example in "Extending stock Docksal images" docs


lmakarov:
https://docs.docksal.io/en/master/advanced/extend-images/#extending-stock-docksal-images

```yaml
# Use a stock Docksal image as the base
FROM docksal/web:2.1-apache2.4

# Install addtional apt packages
RUN apt-get update && apt-get -y --no-install-recommends install \
    libapache2-mod-proxy-html \
    # Cleanup
    && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*

RUN \
    # Enable additional Apache modules
    a2enmod proxy_html proxy_http proxy_balancer
```

The example above is no longer relevant for 2.0 web images, which are Alpine Linux based and not Debian.

PR #544, opened 2018-05-22
Updated example of extending docker images for web

sean-e-dietrich:
Solves Issue #517 by updating the documentation to include an update on how to extend the `web` image as it's now built on alpine.


achekulaev:
@lmakarov fyi

PR #549, opened 2018-05-23
Automatically add ssh keys defined in docksal.env

sean-e-dietrich:
Added ability for SSH Keys to be defined within `.docksal/docksal.env` file prefixed with `SECRET_SSH_KEY_` which loops through the variables and grabs the files from the `~/.ssh/` folder.

Additionally I updated the documentation to reflect the new change as well.

Also added additional tests to the confirm the feature is working.


bin/fin
```diff
@@ -2861,6 +2864,41 @@ ssh_add ()
 	return "$ret"
 }
 
+# Loop through and add defined ssh keys
+_add_ssh_keys_defined () {
```
achekulaev: Let us not separate it out into a function. It's logically the same ssh-add operation and functionally will never be re-used outside of ssh-add

bin/fin
```diff
@@ -2844,6 +2844,30 @@ ssh_add ()
...
+			if is_tty; then
+				# Run ssh-add interactively to allow entering a passphrase for ssh keys (if set)
+				${winpty} docker exec -it docksal-ssh-agent /run.sh ssh-add "${ssh_key_name}"
+			else
+				# In a non-interactive environment (e.g. CI) run non-interactively. Keys with passphrases cannot be used!
+				echo-yellow "Running in a non-interactive environment. SSH keys with passphrases cannot be used."
+				docker exec docksal-ssh-agent /run.sh ssh-add "${ssh_key_name}"
+			fi
+			local ret=$?
+			[[ ! $? -eq 0 ]] && return $ret
```
achekulaev: Should not return here

bin/fin
```diff
@@ -2844,6 +2844,30 @@ ssh_add ()
 	# Only do this when given an argument (key name) or nothing, but not for options (e.g. -l)
 	[[ "${1#-}" == "$1" ]] && docker cp "$ssh_path" docksal-ssh-agent:/
 
+	if [[ "${1#-}" == "$1" ]]; then
```
achekulaev: We should improve this check. A key can have a dashes in it's name (`acquia-key`) and this condition removes dashes in a string and compares to its original will be false.

```bash
if [[ ! "${1}" =~ ^- ]]; then
```

achekulaev:
Please update test to match removed echo

PR #553, opened 2018-05-24
ngrok share fix

sean-e-dietrich:
When supplying arguments for Ngrok I am getting the following issue:

```
docker: Error response from daemon: oci runtime error: container_linux.go:265: starting container process caused "exec: \"ngrok http -authtoken=\\\"XXX\\\"  testdrupal_web_1.testdrupal_default:80\": executable file not found in $PATH".
```

This was due to the following being set in my `docksal.env`

```
NGROK_HOSTNAME="testdrupal.docksal"
NGROK_AUTH="XXX"
```

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -4643,7 +4643,7 @@ ngrok_share ()
 		--link ${container_name} \
 		--name ${ngrok_container_name} \
 		wernight/ngrok \
-		"$ARGS"
+		$ARGS
 }
 
 diagnose ()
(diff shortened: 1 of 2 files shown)


docs/tools/ngrok.md
```diff
@@ -66,3 +66,4 @@ NGROK_PROTOCOL | Can either be `HTTP` or `TCP`, and it defaults to `HTTP` if not
 NGROK_PORT | Port to expose (defaults to `80` for `HTTP` protocol).
 NGROK_REGION | Location of the ngrok tunnel server; can be `us` (United States, default), `eu` (Europe), `ap` (Asia/Pacific) or `au` (Australia)
 NGROK_DEBUG | To debug the connection and see a more detailed log set this to 1
+NGROK_HEADER | Set Host Header. Set to the domain that should be past through to the host.
```
shelane: Typo "past" should be "passed" in this sentence.

bin/fin
```diff
@@ -4643,7 +4643,7 @@ ngrok_share ()
 		--link ${container_name} \
 		--name ${ngrok_container_name} \
 		wernight/ngrok \
-		"$ARGS"
+		$ARGS
```
achekulaev: Why was this changed?

sean-e-dietrich: @achekulaev The `$ARGS` is wrapped in quotes which is causing issues with duplicate quoting. The args are meant to be sent separately. Do you think that with or without the quotes it shouldn't matter? Also the arguments have spaces between them when need to be sent as arguments to the `ngrok` command.

achekulaev: Ah right. Quoting them signals bash it's a single value. OK. Thanks

PR #556, opened 2018-05-25
Documentation updates

shelane:
Usage of "e.g." is always followed by a comma and is preceded with a comma or placed in parentheses. It should not start a sentence as it introduces an apposition in a sentence.

Checked consistency of usage for no periods at the end of bullet points, semicolons with "run" before commands, and the oxford comma placement.

The oxford comma (the comma before the final item of a list) is recommended for readability as it removes any ambiguity. This is especially true where a list has both an "and" and an "or" in it.


lmakarov:
Thanks, @shelane!
What do you think about proofreading https://blog.docksal.io? :)

shelane:
Does the blog need proofreading? I’m far less picky in blogs than in documentation ;-)

PR #558, opened 2018-05-26
Updated options for projects.

sean-e-dietrich:
Updated sample projects list in the docs

### DIFF
--- a/docs/getting-started/setup.md
+++ b/docs/getting-started/setup.md
@@ -24,14 +24,23 @@ fin init
 
 Try one of the example projects:
 
-- [Drupal 7 sample project](https://github.com/docksal/drupal7)  
+### PHP Based
 - [Drupal 8 sample project](https://github.com/docksal/drupal8)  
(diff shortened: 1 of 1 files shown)


lmakarov:
Let's have this list, the one in `README.md` and `fin project create` match.
Also, @achekulaev was proposing to change the naming convention for all example repos to `docksal/boilerplate-<project>` and refer to them as boilerplates instead of examples. That would require renaming the all relevant repos and updating URLs in `fin`.

@achekulaev please post here when you're done updating the hash salt security considerations on the `docksal/drupal8` repo. After that we can proceed with renaming.

sean-e-dietrich:
@lmakarov @achekulaev I like the idea of renaming them, my only question is that if you rename them are they going to throw errors when doing git clones on the old repos?

lmakarov:
@sean-e-dietrich no, Github sets up redirects automatically, which work across the board (web/git and https/ssh).

PR #560, opened 2018-05-27
Fixed non tty message showing in tty environment

shelane:
I was receiving the message in fin 1.59.0 "Running in a non-interactive environment. SSH keys with passphrases cannot be used." No other non-tty messages were showing. I found that this function was asking if it was in a tty environment but showing a message to the contrary. Just added the ! to the check.


achekulaev:
Thanks Shelane

PR #563, opened 2018-05-30
fixed missing hosts file path on WSL

svict4:
`HOSTS_FILE` path was empty when on WSL in `hosts ()` thus `hosts_add|hosts_remove ()` would fail


achekulaev:
Thanks @svict4

Issue #487, opened 2018-03-30
Document how to xdebug docksal drupal with atom


rosenstrauch:
**Description**

_xdebug with atom Works exactly as it should_

still took me a while to figure out that it does
for anyone else maybe this can save some time when added to the documentation

**Steps to reproduce the issue:**

1. apm install php-debug
2. enable Xdebug in docksal-local.env
3. add the following to .docksal/etc/php/php.ini and  .docksal/etc/php/php-cli.ini so the server knows how to connect to the ide
```
xdebug.remote_enable=1
xdebug.remote_host=192.168.64.1
; xdebug.remote_connect_back=1    # Not safe for production servers
xdebug.remote_port=9000
xdebug.remote_handler=dbgp
xdebug.remote_mode=req
xdebug.remote_autostart=true
xdebug.idekey=xdebug.atom
```
in atom enter in the path maps so ide knows where on the server the files are
`/var/www/;/full/path/to/project/`

**Describe the results you received:**

Able to set breakpoints including in behat contexts

**Describe the results you expected:**

Able to set breakpoints including in behat contexts

**Actionables:**
make pr against the docs

lmakarov:
Addressed via #569

PR #569, opened 2018-05-31
Added Atom Xdebug docs

sean-e-dietrich:
Added docs for configuring and using Atom and PHP XDebug with Docksal

Solves #487


achekulaev:
👍

Issue #584, opened 2018-06-05
fin config get/set --env


lmakarov:
Extend `fin config get/set` with support for specific environment configurations.

```
fin config set --env=<environment> VARIABLE=VALUE
```

The above will sets the variable in `.docksal/docksal-<environment>.env`

PR #596, opened 2018-06-08
added env option for fin config set/remove

sean-e-dietrich:
Added option for `fin config set/remove` to pass an `--env` flag.

Example of command

```
$ fin config set --env=local TEST=1
Added value for TEST into HOME/.docksal/docksal.env
```

Additionally added tests and updated help command to reflect the changes.

Solves #584

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1686,13 +1686,16 @@ show_help_config ()
 	printh "    --docroot=mydir" "Set non-default DOCROOT during config generate"
 	printh "set [options] [VAR=VAL]" "Set value(s) for the variable(s) in project ENV file"
 	printh "    --global" "Set for global ENV file"
+	printh "    --env=[name]" "Set in environment specific project ENV file (default: local)"
 	echo
 	printh "remove [options] [VAR]" "Remove variable(s) from project ENV file"
 	printh "rm [options] [VAR]" ""
 	printh "    --global" "Remove from global ENV file"
+	printh "    --env=[name]" "Remove from environment specific project ENV file (default: local)"
 	echo
 	printh "get [options] [VAR]" "Get the value of the single variable from project ENV file"
 	printh "    --global" "Get value from global ENV file"
+	printh "    --env=[name]" "Get value from environment specific project ENV file (default: local)"
 
 	echo
 	echo "Examples:"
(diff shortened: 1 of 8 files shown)


bin/fin
```diff
@@ -5146,6 +5148,10 @@ config_get ()
 		target_env_file="$CONFIG_ENV"
 	else
 		load_configuration
+		if [[ -n $env ]]; then
+			shift
```
achekulaev: shift is not needed here

achekulaev: First I wanted to write that shift is not needed, but then I understood why it is here. 
It was my mistake. 
We should use ${ARGV[0]} instead of $1 in line 5158, because parse_arguments have already parsed it into that array.
And then this shift will not be needed here

bin/fin
```diff
@@ -5162,6 +5168,12 @@ config_set ()
 		target_env_file="$CONFIG_ENV"
 	else
 		load_configuration
+		if [[ -n $env ]]; then
+			local docksal_env_file="$(get_project_path_dc)/.docksal/docksal-${env}.env"
+			if [[ ! -f $dockal_env_file ]]; then
+				touch $docksal_env_file
```
achekulaev: 1. Always use quotes around paths
2. Error processing

```bash
touch "$docksal_env_file"
if_failed_error "Could not create $docksal_env_file"
```

achekulaev:
This PR now contains a lot of unrelated changes to indentation many of which are actually wrong

sean-e-dietrich:
@achekulaev noticed that when I adjusted the code in PHPStorm it reset some of the spacing on items. I reset the commits and am now starting back from where I was before without those spaces.

bin/fin
```diff
@@ -5206,23 +5225,34 @@ config_remove ()
...
+	if [[ "${#ARGV[@]}" == 0 ]]; then
+		echo "No values to remove"
+		return 1
+	fi
+
+	# Loop named params array
+	for param in "${ARGV[@]}"
 	do
-		if [[ "$1" == "--global" ]]; then
+		if [[ "${param}" =~ ^- ]]; then
```
achekulaev: this condition is not needed

sean-e-dietrich: @achekulaev what ends up happening is if I use `--env` it pics it up as a variable. If that is the case then I believe that `parse_params` may need to be looked at because when we set `--env=local` without this piece `$ARGV` is adding `--env` to the variable.

achekulaev:
```bash
§ fin debug parse_params --env=local TEST
local ARGV=(); local ARGN=(); local ARGO=();
ARGN+=('--env=local');
local env="local";
ARGV+=('TEST');
```

everything seems to be in order, `--env` does not get into ARGV.

sean-e-dietrich: @achekulaev you are correct, I was getting the config_remove and the config_set module mixed up.

PR #597, opened 2018-06-09
Feature/documentation

shelane:
Some more punctuation, wording, etc.


shelane:
@sean-e-dietrich not at all. I just happen to have an eye that won’t let it go. I’m happy to be able to contribute to this project in any way I can.

Issue #590, opened 2018-06-07
Is there a way to set docroot to current directory?


goldingdamien:
If I have say my_project/.docksal and all the code for my_project starts in my_project say my_project/index.php, can I change the config so that my_project becomes docroot instead, and if so how and are there any side effects?

If not, my desire is to have an easy way to start a LAMP environment without moving any project files such as possible in http-server in npm(node js). Any alternatives would be considered. Regards.
* I am new to docksal so may have missed something.

sean-e-dietrich:
Hi @goldingdamien yes it is possible.

In `.docksal/docksal.env` set the following:

```
DOCROOT=.
```

This will tell Docksal that your current document root is the same as your project root. Once you have done that a `fin restart` should be all that is necessary for Docksal to pick up the changes.

goldingdamien:
Works fine, thanks.

PR #599, opened 2018-06-09
Added tip to project setup for docroot

sean-e-dietrich:
Added documentation to Project Setup so so that users know they can use the current directory as the `DOCROOT`

Solves #590


shelane:
If so, then...

shelane:
I was actually trying to find the place in the docs that discuss setting the docroot as I was thinking that would be a logical place for it to go.

PR #611, opened 2018-06-13
Support ngrok configuration file for `fin share`

japerry:
While adding new options is a great feature, having an ngrok configuration file gives you full access to the power of ngrok.

This PR adds the ability to use a configuration file, but falls back to the environment parameters if its not set.

It'd be nice if you could override the config file with cli options, but in my testing that didn't quite work for me, which is why the code leaves the existing launch command intact and overrides that only when the file exists.

Also in the PR is an example ngrok file for folks to try it out.


sean-e-dietrich:
@japerry environment variables can be set within `.docksal-local.env` and then used for reference with ngrok. But, I could see what you are talking about. If you use the config file doe the other flags get ignored?

achekulaev:
Hi @japerry and thanks for contribution. I like the idea. 
I have 2 comments on standards. 
1. Please make filename `.docksal/etc/ngrok/ngrok.yml`
2. Please add the description of this option to `docs/tools/ngrok.md`

bin/fin
```diff
@@ -4805,13 +4805,26 @@ ngrok_share ()
...
+    # Load a configuration if its available
+    ngrok_conf="$(get_project_path_dc)/.docksal/ngrok.local.yml"
+    if [[ -f "$ngrok_conf" ]]; then
+        ${winpty} docker run --rm -v $(get_project_path_dc):/var/www -it \
+            --publish 4040 \
+            --net ${network} \
+            --link ${container_name} \
+            --name ${ngrok_container_name} \
+            wernight/ngrok \
+            sh -c "ngrok start -config /var/www/.docksal/ngrok.yml -all"
```
achekulaev: Instead of code duplication `$ARGS` variable should be overridden

```bash
  # Load a configuration if its available
  ngrok_config="$(get_project_path_dc)/.docksal/ngrok.local.yml"
  if [[ -f "$ngrok_conf" ]]; then
    ARGS='sh -c "ngrok start -config /var/www/.docksal/ngrok.yml -all"'
    NGROK_VOLUME="-v $(get_project_path_dc)/.docksal:/var/www/.docksal"
  fi
  
  ${winpty} docker run -rm ${NGROK_VOLUME} -it...
```

Also note that you used different file name in different lines `ngrok.local.yml` and `ngrok.yml"`

japerry:
The primary reason I'm using a config file is because I want to share out multiple sites (Drupal multisite) with one project. You cannot share two sites otherwise from the same project. The command line arguments don't allow for this. 

I've made some changes to the PR which includes being able to add a local and project based ngrok.yml file.

bin/fin
```diff
@@ -4803,15 +4803,28 @@ ngrok_share ()
 		ARGS="$ARGS -log stdout"
 	fi
 
-	ARGS="$ARGS ${container_name}.${network}:${NGROK_PORT}"
-
-	${winpty} docker run --rm -it \
+	# Overwrite settings with a configuration if its available
+    NGROK_VOLUME="-v $(get_project_path_dc)/.docksal:/var/www/.docksal"
```
achekulaev: Indentation is wrong. Should use tabs

bin/fin
```diff
@@ -4803,15 +4803,29 @@ ngrok_share ()
 		ARGS="$ARGS -log stdout"
 	fi
 
-	ARGS="$ARGS ${container_name}.${network}:${NGROK_PORT}"
-
-	${winpty} docker run --rm -it \
+	local ngrok_conf="$(get_project_path_dc)/.docksal/etc/ngrok/ngrok.yml"
```
```
local ngrok_conf_path="$(get_project_path_dc)/.docksal/etc/ngrok"
local ngrok_conf="${ngrok_conf_path}/ngrok.yml"
...
local local_ngrok_conf="${ngrok_conf_path}//ngrok-${DOCKSAL_ENVIRONMENT}.yml"
```

achekulaev:
Thanks @sean-e-dietrich

Issue #609, opened 2018-06-13
Improve `fin exec` to work with different containers 


achekulaev:
`CONTAINER_NAME` env variable can be defined to tell `fin exec` to execute in a different container.

There should be a better way. 

> IIRC `CONTAINER_NAME ` was mostly an internal workaround in fin.
If we are going to expose it, then I suggest we move this feature into a parameter to `fin exec`, e.g.:
    ```
    fin exec --service=web uname -a
    ```

achekulaev:
Done in other tickets `fin exec --in=db mysql -uroot -p`

PR #615, opened 2018-06-14
Allow for Container Names option with fin exec [WIP]

sean-e-dietrich:
Add ability for containers to use `--container` option when running `fin exec`

Solves https://github.com/docksal/docksal/issues/609

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1323,7 +1323,7 @@ show_help_exec ()
 	printh "-T" "Disable pseudo-tty allocation."
 	printh "" "Useful for non-interactive commands when output is saved into a variable for further comparison."
 	printh "" "In a TTY mode the output may contain unexpected control symbols (garbage)."
-
+    printh "--container=db" "Name of the container to execute command in."
 	echo
 	echo "Advanced:"
 	printh "CONTAINER_NAME=[name]" "Environment variable. Defines container to run in (default is \`cli\`)"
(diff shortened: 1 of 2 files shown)


achekulaev:
No, its a complicated matter and parse params is no friend here. It's for simple parameters parsing. It will fail for parsing things like

`fin exec 'ls -la 2>/dev/null | grep test'`

sean-e-dietrich:
@achekulaev please look again at the changes.

I've also tested the line that you provided. Had to modify for the project I was working with and

```
➜  aes-annualconf git:(master) ✗ fin-develop exec 'ls -la 2>/dev/null | grep composer'
-rw-r--r--   1 docker dialout     5598 Jun 15 00:15 composer.json
-rw-r--r--   1 docker dialout   348047 Jun 15 00:16 composer.lock
➜  aes-annualconf git:(master) ✗ fin-develop exec --container=db 'ls -la 2>/dev/null | grep composer'
➜  aes-annualconf git:(master) ✗ fin-develop exec --container=web 'ls -la 2>/dev/null | grep composer'
-rw-r--r--    1 503      dialout       5598 Jun 15 00:15 composer.json
-rw-r--r--    1 503      dialout     348047 Jun 15 00:16 composer.lock
```

This doesn't fail. I'm specifically using `parse_params` to tell me if any options were used. Converted instead of checking `$1` to see if it's `-T` I'm just saying `$T` and then your `shift` was pushing it onto the next variable. Next I check to see if `--container` exists and is set if so then set it to the `CONTAINER_NAME` variable and then `shift` the argument off the stack. Lastly keeping the original method for reading in the arguments to pass.

If you had a specific vision for this then cool!

achekulaev:
Right, I forgot that the approach was already tried out with `-T` option. 
The only thing then I would change is the name to `--in`, just because it's so long to type `--container` :)

PR #526, opened 2018-05-12
Namespacing of Commands

sean-e-dietrich:
This is to solve the following feature pull request: https://github.com/docksal/docksal/pull/461

PR #618, opened 2018-06-15
Hotfix: fin help doesn't show Custom commands anymore

fjmk:
In #526 are the find commands changed. At least in Ubuntu I don't see my Custom commands since fin v1.90.

In [find manpage](http://man7.org/linux/man-pages/man1/find.1.html) I find the text:
> -perm +mode
>              This is no longer supported (and has been deprecated since
>              2005).  Use -perm /mode instead.

The idea is to list only executable commands. I have opted to use -executable instead of -perm.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1240,11 +1240,16 @@ show_help_list_commands_in_help ()
 	else
 		addons_path="$HOME/$DOCKSAL_COMMANDS_PATH"
 	fi
-
-	local addons_list=$(find "$addons_path" -type f -executable 2>/dev/null)
+	
+	local executable="-executable"
+	if [[ is_mac ]]; then
+		executable="-perm +100"
+	fi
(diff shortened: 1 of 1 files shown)


lmakarov:
@fjmk thanks for the PR!

@sean-e-dietrich we should add a test case for #526 and custom commands.

bin/fin
```diff
@@ -1241,10 +1241,10 @@ show_help_list_commands_in_help ()
 		addons_path="$HOME/$DOCKSAL_COMMANDS_PATH"
 	fi
 
-	local addons_list=$(find "$addons_path" -type f -perm +100 2>/dev/null)
+	local addons_list=$(find "$addons_path" -type f -executable 2>/dev/null)
```
achekulaev: doesnt work on mac

PR #619, opened 2018-06-15
Fin Commands Test and Fix

sean-e-dietrich:
This is a rewrite of the PR https://github.com/docksal/docksal/pull/618

This includes a test for #526

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1242,14 +1242,12 @@ show_help_list_commands_in_help ()
 	fi
 	
 	local executable="-executable"
-	if [[ is_mac ]]; then
-		executable="-perm +100"
-	fi
+	[[ is_mac ]] && executable="-perm +100"
 	
 	local addons_list=$(find "$addons_path" -type f ${executable} 2>/dev/null)
-	if [ ! -z "$addons_list" ]; then
+	if [[ "$addons_list" != "" ]]; then
 		local scope
-		for cmd_name in $(find "$addons_path" -type f ${executable})
+		for cmd_name in ${addons_list}
 		do
 			local g
 			[[ "$1" == 'global' ]] && g=' [g]'
--- a/bin/fin
+++ b/bin/fin
@@ -1240,11 +1240,16 @@ show_help_list_commands_in_help ()
 	else
 		addons_path="$HOME/$DOCKSAL_COMMANDS_PATH"
 	fi
-
-	local addons_list=$(find "$addons_path" -type f -executable 2>/dev/null)
+	
+	local executable="-executable"
+	if [[ is_mac ]]; then
+		executable="-perm +100"
+	fi
+	
+	local addons_list=$(find "$addons_path" -type f ${executable} 2>/dev/null)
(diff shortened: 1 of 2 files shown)


sean-e-dietrich:
This is an extension of the following PR but with tests https://github.com/docksal/docksal/pull/618

bin/fin
```diff
@@ -1240,11 +1240,16 @@ show_help_list_commands_in_help ()
 	else
 		addons_path="$HOME/$DOCKSAL_COMMANDS_PATH"
 	fi
-
-	local addons_list=$(find "$addons_path" -type f -perm +100 2>/dev/null)
+	
+	local executable="-executable"
+	if [[ is_mac ]]; then
```
lmakarov: This can be a one-liner: `is_mac  && executable="-perm +100"`

bin/fin
```diff
@@ -1240,11 +1240,16 @@ show_help_list_commands_in_help ()
...
+		executable="-perm +100"
+	fi
+	
+	local addons_list=$(find "$addons_path" -type f ${executable} 2>/dev/null)
 	if [ ! -z "$addons_list" ]; then
 		local scope
-		for cmd_name in $(find "$addons_path" -type f -perm +100)
+		for cmd_name in $(find "$addons_path" -type f ${executable})
```
lmakarov: `for cmd_name in ${addons_list}`?

bin/fin
```diff
@@ -1240,11 +1240,16 @@ show_help_list_commands_in_help ()
...
-
-	local addons_list=$(find "$addons_path" -type f -perm +100 2>/dev/null)
+	
+	local executable="-executable"
+	if [[ is_mac ]]; then
+		executable="-perm +100"
+	fi
+	
+	local addons_list=$(find "$addons_path" -type f ${executable} 2>/dev/null)
 	if [ ! -z "$addons_list" ]; then
```
lmakarov: We are using bash style tests instead of sh, so `if [[ "$addons_list" != "" ]]; then` here

tests/smoke-test-general.bats
```diff
@@ -62,6 +62,22 @@ SKIP=1
 	unset output
 }
 
+@test "fin team/test" {
+	[[ $SKIP == 1 ]] && skip
+
+	cmd='#!/bin/bash
```
lmakarov: Use heredoc:

```
    cat <<EOF > .docksal/commands/team/test
#!/bin/bash
echo "Test Command"
EOF
```

lmakarov: Looks like this needs to be applied in several other places in this file

tests/smoke-test-general.bats
```diff
@@ -62,6 +62,22 @@ SKIP=1
 	unset output
 }
 
+@test "fin team/test" {
```
lmakarov: Let's use `namespace/custom-command` to make it more descriptive

achekulaev:
Thanks Sean 👍

PR #642, opened 2018-06-23
Platform.sh Integration

sean-e-dietrich:
### Pending PR Approval: https://github.com/docksal/service-cli/pull/54

* Added docs for Platform.sh Usage
* Added command `fin platform ...`
* Added help text for `platform` command


docs/tools/platformsh.md
```diff
@@ -0,0 +1,27 @@
+# Platform.sh
+
+Platform.sh offers a command line tool to help . It enables you to do almost everything in a terminal that you can do in the Dashboard, and much more.
```
shelane: command line too to help... ? Looks like it's missing something. To help development for platform.sh?

The text for Terminus states:
>Terminus is Pantheon's command line tool. It enables you to do almost everything in a terminal that you can do in the Dashboard, and much more.

docs/tools/platformsh.md
```diff
@@ -0,0 +1,27 @@
...
+
+Platform.sh's CLI is pre-installed in the `cli` container. Note: required `docksal/cli` version: 2.3.0+.
+
+
+## Configuration
+
+Follow the [official docs](https://docs.platform.sh/gettingstarted/cli/api-tokens.html) to generate your API token.
+
+Add the token to `$HOME/docksal/docksal.env`:
```
lmakarov: Rewrite to use `fin config set --global`. May still mention that the variable will be stored in `$HOME/docksal/docksal.env`

docs/tools/platformsh.md
```diff
@@ -0,0 +1,27 @@
...
+```
+
+Refresh `cli` service configuration with `fin project start`. This will inject the token into `cli` and authenticate
+the container with Pantheon.
+
+Use `fin platform <command>` from the host or `platform <command>` inside `cli`.
+
+Please refer to the [official docs](https://docs.platform.sh/gettingstarted/cli.html#usage) for usage details.
+
+Note: It is also possible to add/override these values via `.docksal/docksal.env` and `.docksal/docksal-local.env` at the project level. Keep in mind, `.docksal/docksal.env` is a shared configuration file and should be committed to git. `.docksal/docksal-local.env`, on the other hand, can be used for local overrides and should be excluded from git. As such, it is best to use `.docksal/docksal-local.env` for any personal keys and tokens configured at the project level.
```
lmakarov: Same here - let's standardize around `fin config set` moving forward for configuration adjustments. Should still mention the `docksal-local.env` (`fin config set --env=local`) use case.

shelane: Is this a Pantheon service? Line 25 says "with Pantheon."

sean-e-dietrich: thanks @shelane I removed the reference of Pantheon

PR #643, opened 2018-06-24
fin run-cli documumentation update

fjmk:
Two changes:
- ```fin compose .....``` should be ```fin run-cli compose .....``` 
- Add DOCROOT=web to .docksal.docksal.env


achekulaev:
Thanks!

https://github.com/docksal/docksal/pull/645/files one had more fixes including fixes in this PR, so I merged that one.

PR #651, opened 2018-06-27
Composer command

sean-e-dietrich:
Added command to make running composer commands easier. Also to make it known that Composer is a tool that is installed already within the container.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1207,6 +1207,7 @@ show_help ()
 	printh "drupal [command]" "Drupal Console command (requires Drupal 8)"
 	printh "terminus [command]" "Pantheon's Terminus (requires docksal/cli 2.1+)"
 	printh "wp [command]" "WordPress CLI command (requires WordPress)"
+	printh "composer [command]" "Run Composer commands"
 
 	echo
 	printh "addon <command>" "Addons management commands: install, remove (${yellow}fin help addon${NC})"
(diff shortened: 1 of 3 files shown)


docs/tools/composer.md
```diff
@@ -0,0 +1,24 @@
...
+From with the cli container (`fin bash`) composer can be called directly:
+
+```
+composer --version
+```
+
+!!! note "Composer must be run inside the same directory as composer.json"
+    Unless the the `--working-dir=` is used composer must be ran within the same directory or a child of that directory
+    as the projects composer.json
+
```
shelane: "used, the `composer` command must be run within the directory or child directory containing the project's composer.json file."

bin/fin
```diff
@@ -6292,6 +6293,15 @@ case "$1" in
 			_exec wp "$@"
 		fi
 		;;
+	composer)
+		load_configuration
+		shift
+		if [[ "$1" == "" ]]; then
+			_exec composer
```
achekulaev: I guess there is no need in this for composer. Just do `_exec composer "$@"` all the time.
It was needed for drush for drush-specific reasons

bin/fin
```diff
@@ -6292,6 +6293,11 @@ case "$1" in
 			_exec wp "$@"
 		fi
 		;;
+	composer)
+		load_configuration
```
achekulaev: Sorry, missed it. There is no need to load configuration. Actually it is counterproductive o do it, because it will throw in case it does not find a project.

sean-e-dietrich: @achekulaev I that is the main purpose to run similar to drush and wp-cli where it runs inside of the projects container therefore it needs to make sure the project exists before hand.

achekulaev: Right. My bad.

Issue #430, opened 2017-12-23
Add verification for and document characters allowed in VIRTUAL_HOST 


frederickjh:
**Description**

I am posting this here from gitter.im chat as requested by @lmakarov. See chat text below included for reference.

Currently there is no documentation or validation for allowed characters in VIRTUAL_HOST. Possible tasks to fix this are:
* Add documentation including the Wikipedia link regarding domain names and the allowed characters.
* Add validation to the `fin config show` command that would give an error that the configuration is invalid.
* Add validation when `fin up` is run that would show a warning that the VIRTUAL_HOST is invalid and that Docksal DNS routing will not work.

----

**From [gitter.im/docksal/community-support](https://gitter.im/docksal/community-support)**
#### Chris Zietlow @aczietlow Dec. 01 18:02
I’ve run into a weird issue with the proxy service. I’ve binded the service to 0.0.0.0 so that it handles all incoming requests, and it works as expected. Except when there is an _ in the url
~/.docksal/docksal.env
```
DOCKSAL_VHOST_PROXY_IP=“0.0.0.0”
```
project docksal.env
```
COMPOSE_PROJECT_NAME=Pied_Piper-features-huli-chat
VIRTUAL_HOST=features-huli-chat.Pied_Piper.mg.zietlow.ninja
```
Everything works until an **_** gets added

#### Chris Zietlow @aczietlow Dec. 01 18:08
Is this a given with proxy services, or a bug?
i.e. Is this to be expected, or would it be helpful for me to dig further into it?

#### Leonid Makarov @lmakarov Dec. 05 01:01
> Except when there is an _ in the url

@aczietlow you cannot use underscores in a URL

From https://en.wikipedia.org/wiki/Domain_name

> Domain names may be formed from the set of alphanumeric ASCII characters (a-z, A-Z, 0-9)

also, you cannot have uppercase letters in VIRTUAL_HOST - these won't be routable by the proxy, since browsers force lowercase domain names
It would be helpful to post these issues on Github, so others can easier discover them

achekulaev:
@frederickjh I actually wrote a post about this after that case blog.docksal.io/Virtual-host-value-limitations-4f105725dbf8
But yeah some automation would not hurt

lmakarov:
[...]

PR #653, opened 2018-06-27
Add a check in place, so that virtual hosts comply with browser standards

sean-e-dietrich:
* Added warning into place to let user know that the VIRTUAL_HOST has been modified.
* Converted virtual host to all lower case
* Converted underscores to dashes
* Added test to make sure this is happening
* Additionally removed all characters that aren't a `-` or `0-9` or `a-z` or `.`

Should Solve #430

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -5210,7 +5210,7 @@ load_configuration ()
 
 	if [[ $(echo "${VIRTUAL_HOST}" | grep -c -v "^[a-z0-9\-\.]*$") -gt 0 ]]; then
 		local OLD_VHOST=${VIRTUAL_HOST}
-		export VIRTUAL_HOST=$(echo ${VIRTUAL_HOST} | tr '[:upper:]' '[:lower:]' | sed 's/_/-/g')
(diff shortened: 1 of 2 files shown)


bin/fin
```diff
@@ -5208,6 +5208,12 @@ load_configuration ()
 	fi
 	export DOCROOT=${DOCROOT:-docroot}
 
+	if [[ $(echo "${VIRTUAL_HOST}" | grep -c -v "^[a-z0-9\-\.]*$") -gt 0 ]]; then
```
achekulaev: I have a refactoring proposition

```bash
local OLD_VHOST=${VIRTUAL_HOST}
export VIRTUAL_HOST=$(echo ${VIRTUAL_HOST} | tr '[:upper:]' '[:lower:]' | sed 's/_/-/g' | sed 's/[^-a-z0-9\.]//g')
# Warn in case Virtual Host was auto-corrected
if [[ "$OLD_VHOST" != "$VIRTUAL_HOST" ]]; then
    echo-warning "The ${yellow_bold}VIRTUAL_HOST${yellow} has been modified from ${yellow_bold}${OLD_VHOST}${yellow} to ${yellow_bold}${VIRTUAL_HOST}${NC} ${yellow}to comply with browser standards."
fi
```

Issue #625, opened 2018-06-20
Docksal OpenSSL?


loopy3025:
Our network here at my job has changed rather drastically of late and I've finally been able to get a Drupal 7 and Drupal 8 site running on my local dev using both http and https. I am ecstatic! There's just one more roadblock and that is downloading and enabling modules. 

I run
```
$ frush en examples -y
```
and I get
```
examples was not found.                                                                                                            [warning]
simplexml_load_file(): I/O warning : failed to load external entity "" Project.php:74                                              [warning]
Failed to get available update data from https://updates.drupal.org/release-history/examples/8.x  
```
I googled a bunch and found this telling me to enable the open.ssl extension in php: https://www.drupal.org/project/infrastructure/issues/2661298

And this discussing enabling php modules:
https://github.com/docksal/service-cli/issues/9

When I tried adding this to the site's .docksal\etc\php\php.ini file, drupal got very, very angry with me when I ran cache-rebuild.
```
extension_dir=C:/PHP7/ext
extension=php_openssl.dll
```
I'm wondering if all that is even necessary, though.

**Output of `fin config`:**

<details>
  <summary>fin config output</summary>

```
---------------------
COMPOSE_PROJECT_NAME_SAFE: examples
COMPOSE_FILE:
C:/Users/afasoldt/.babun/cygwin/home/afasoldt/.docksal/stacks/volumes-bind.yml
C:/Users/afasoldt/.babun/cygwin/home/afasoldt/.docksal/stacks/stack-default.yml
C:/Users/afasoldt/.babun/cygwin/home/afasoldt/projects/examples/.docksal/docksal.yml
ENV_FILE:
C:/Users/afasoldt/.babun/cygwin/home/afasoldt/projects/examples/.docksal/docksal.env
C:/Users/afasoldt/.babun/cygwin/home/afasoldt/projects/examples/.docksal/docksal-local.env

PROJECT_ROOT: /c/Users/afasoldt/.babun/cygwin/home/afasoldt/projects/examples
DOCROOT: docroot
VIRTUAL_HOST: examples.docksal
VIRTUAL_HOST_ALIASES: *.examples.docksal
IP: 192.168.64.100
MYSQL: 192.168.64.100:32771

Docker Compose configuration
---------------------
services:
  cli:
    dns:
    - 192.168.64.100
    - 8.8.8.8
    environment:
      BLACKFIRE_CLIENT_ID: null
      BLACKFIRE_CLIENT_TOKEN: null
[...]

PR #663, opened 2018-06-29
Auto-detect upstream DNS when possible

### LINKED ISSUE #586: Automatically switch to local DNS when Google DNS blocked
A lot of people in corporate networks meet the DNS issue and it deters many newcomers from using Docksal.
It would be beneficial it we could automate the Upstream DNS switch.

On macOS and Linux current DNS server can be found in `/etc/resolv.conf`
On Windows it looks like Primary WINS Server might be the server we are looking for:

```bash
ipconfig /all | grep "Primary WINS"
```

We might also switch it back to `8.8.8.8` in case we detect that DOCKSAL UPSTREAM DNS is overridden but not accessible while `8.8.8.8` is accessible (person brought laptop home).

achekulaev:
Closes #586

FYI @sean-e-dietrich

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -3243,9 +3243,9 @@ detect_dns ()
 		nameservers=$(grep "nameserver" "/etc/resolv.conf" 2>/dev/null | sed "s/nameserver //")
 		# Find first suitable
 		if [[ "$nameservers" != "" ]]; then
-			for server in "$nameservers"; do
+			for server in $nameservers; do
 				# Check that it is an IP address and exclude Docksal IP
-				if [[ "$server" =~ [1-9]{1,3}\.[1-9]{1,3}\.[1-9]{1,3}\.[1-9]{1,3} ]] &&
+				if [[ "$server" =~ [0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3} ]] &&
 					[[ "$server" != "192.168.64.100" ]]; then
 					DOCKSAL_DNS_UPSTREAM="$server"
 				fi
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -3231,16 +3231,57 @@ install_proxy_service ()
...
+			done
+		else
+			# In case no name servers found, fall back to the default
+			DOCKSAL_DNS_UPSTREAM="$DOCKSAL_DEFAULT_DNS"
+		fi
+	elif is_windows; then
+		# Look for Primary WINS server, it represents DNS whe WINS services is in use
+		local primary_wins=$(ipconfig /all | grep "Primary WINS")
+		if [[ "$primary_wins" =~ ([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}) ]]; then
+			DOCKSAL_DNS_UPSTREAM=${BASH_REMATCH[0]}
```
lmakarov: OMG what does that even do?? :)

bin/fin
```diff
@@ -3231,16 +3231,57 @@ install_proxy_service ()
...
+					DOCKSAL_DNS_UPSTREAM="$server"
+				fi
+			done
+		else
+			# In case no name servers found, fall back to the default
+			DOCKSAL_DNS_UPSTREAM="$DOCKSAL_DEFAULT_DNS"
+		fi
+	elif is_windows; then
+		# Look for Primary WINS server, it represents DNS whe WINS services is in use
+		local primary_wins=$(ipconfig /all | grep "Primary WINS")
```
lmakarov: Can we rely on DNS and WINS being always the same server/IP on every Windows network?

bin/fin
achekulaev: BASH_REMATCH array contains matches from the last regexp matching (line above).

bin/fin
achekulaev: I can't be 100% sure of course. But it so for my local Windows 10 and for a guy in a strict corporate network with DNS banned and certificates-gated internet access, whom I asked to execute it https://github.com/docksal/docksal/issues/625#issuecomment-401441119

I am also asking Sean right now and will post update here.

bin/fin
```diff
@@ -3231,16 +3231,57 @@ install_proxy_service ()
...
+
+	if is_linux || is_mac; then
+		# Get from /etc/resolv.conf
+		nameservers=$(grep "nameserver" "/etc/resolv.conf" 2>/dev/null | sed "s/nameserver //")
+		# Find first suitable
+		if [[ "$nameservers" != "" ]]; then
+			for server in $nameservers; do
+				# Check that it is an IP address and exclude Docksal IP
+				if [[ "$server" =~ [0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3} ]] &&
+					[[ "$server" != "192.168.64.100" ]]; then
```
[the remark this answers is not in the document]
achekulaev: Thanks for reminding. I refactored all those network variables. That was long overdue.

bin/fin
achekulaev: Sean does not have Primary WINS in his configuration. So far so good. When WINS is present it is the DNS, which makes sense.

Issue #174, opened 2017-04-12
Create a drupal-composer/drupal-project based example repo for Drupal 8


lmakarov:

JensvdHeydt:
I've already done something similar. Would you be interested?

achekulaev:
@JensvdHeydt sure

davekopecek:
I've got a `Docksal powered Drupal 8 Composer Installation`  here: git@github.com:davekopecek/drupal8composer.git

It uses a shared volume to persist package manager cache as suggested in  https://github.com/docksal/docksal/issues/231

achekulaev:
@lmakarov drawing your attention that composer based Drupal installation is NOT an official way to install Drupal. 
There is a long comment (sprite commented 9 November 2017 at 17:55) describing differences between that commonly used composer script and vanilla Drupal here: https://www.drupal.org/forum/support/installing-drupal/2017-11-08/where-to-install-drupal-8-with-composer

lmakarov:
I renamed this issues to add more clarity.

Out of the box, Drupal core does ship with [compose.json](https://github.com/drupal/drupal/blob/8.4.x/composer.json) and using Composer IS the official way of managing dependencies in D8. Not so much for installing D8, though.

https://github.com/drupal-composer/drupal-project has a goal of making it "easier" to install D8 using Composer from scratch. It is indeed an unofficial (and opinionated) way on how Drupal codebase should be structured. And yes - it is suited more for advanced developers than novice and general public.

achekulaev:
@lmakarov I have an example of a developer who installed Drupal in Docksal via composer. After the installation he had to downgrade `doctrine/annotations` from 1.5 to 1.4  (`composer require doctrine/annotations:1.4`) as he was having issues otherwise. I don't have details on the nature of the issues he was experiencing, just mentioning it here.

To be even more precise he used Drupal Console rather than plain composer (`drupal site:new`) but it runs the same composer command 
```
composer create-project drupal-composer/drupal-project:8.x-dev docroot --prefer-dist --no-progress --no-interaction
```

lmakarov:
Here's repo structure Acquia recommends for Drupal 8 sites:
https://docs.acquia.com/article/composer-migrate-existing-d8-site-composer-managed-build

And this is their top level composer.json template:
https://github.com/acquia/acquia-ra-composer/blob/master/composer.json

That looks to be less invasive and opinionated compared to `drupal-composer/drupal-project`.

frederickjh:
[...]

PR #668, opened 2018-07-02
Added Drupal8 Composer Repo as list of options

sean-e-dietrich:
Solves #174 

Added option for Drupal8 Composer project within `fin project create` method.


lmakarov:
The name of the folder locally is the name of the repo, right?
So in this case the folder will be named `boilerplate-drupal8-composer` and the URL is going to be http://boilerplate-drupal8-composer.docksal.

I think that looks ugly and we should remove the `boilerplate-` prefix when doing a checkout (or just checkout using a predefined folder name)

sean-e-dietrich:
When we create the folder name we create based on the project name that is supplied as question 1. Therefore the folder name won't be `boilderplate-drupal8-composer`

lmakarov:
k, all clear then :)

PR #670, opened 2018-07-02
Fix fin @alias config get/set/rm

lmakarov:
Fixes an issue when fin config get/set/rm does not work with an alias


achekulaev:
Just stylistic fixes right?

lmakarov:
@achekulaev no, I dropped `load_configuration` in config get/set/rm function, since that does not work when you use a project alias.

The old code relied on `$docksal_env_file` being available after `load_configuration` is triggered.
When fin is used with an alias, that value is lost, since `load_configuration` does not export it.

Issue #674, opened 2018-07-03
ngrok is broken


achekulaev:
```bash
22:26:03 ~/Projects/drupal8
§ fin share
Exposing domain drupal8.docksal via ngrok...
WARN[07-03|02:25:58] failed to get home directory, using $HOME instead err="user: Current not implemented on linux/amd64" $HOME=/home/ngrok
NAME:
   http - start an HTTP tunnel

USAGE:
   ngrok http [command options] [arguments...]

DESCRIPTION:
   Starts a tunnel listening for HTTP/HTTPS traffic with a specific hostname.
   The HTTP Host header on incoming public requests is inspected to
   determine which tunnel it matches.

   HTTPS endpoints terminate TLS traffic at the ngrok server using the
   ngrok.io certificates. The decrypted, HTTP traffic is then forwarded
   through the secure tunnel and then to your local server. If you don't want
   your TLS traffic to terminate at the ngrok server, use a TLS or TCP tunnel.

EXAMPLES:
   ngrok http 8080                      # forward ngrok.io subdomain to port 80
   ngrok http example.com:9000          # forward traffic to example.com:9000
   ngrok http -subdomain=bar 80         # request subdomain name: 'bar.ngrok.io'
   ngrok http -hostname=ex.com 1234     # request tunnel 'ex.com' (DNS CNAME)
   ngrok http -auth='falken:joshua' 80  # enforce basic auth on tunnel endpoint
   ngrok http -host-header=ex.com 80    # rewrite the Host header to 'ex.com'

OPTIONS:
   --auth 		enforce basic auth on tunnel endpoint, 'user:password'
   --authtoken 		ngrok.com authtoken identifying a user
   --bind-tls "both"	listen for http, https or both: true/false/both
   --config		path to config files; they are merged if multiple
   --host-header 	set Host header; if 'rewrite' use local address hostname
   --hostname 		host tunnel on custom hostname (requires DNS CNAME)
   --inspect		enable/disable http introspection
   --log "false"	path to log file, 'stdout', 'stderr' or 'false'
   --log-format "term"	log record format: 'term', 'logfmt', 'json'
   --log-level "info"	logging level
   --region 		ngrok server region [us, eu, au, ap] (default: us)
   --subdomain 		host tunnel on a custom subdomain

ERROR:  You must specify a single argument: a port or address to tunnel to.
ERROR:  You specified 0 arguments: []
ERROR:  For example, to expose port 80, run 'ngrok http 80'.
```

[...]

PR #676, opened 2018-07-03
NGROK Automated Tests

sean-e-dietrich:
The following is to test the basic `fin share` command in light of issue #674

### DIFF
--- a/tests/smoke-test-general.bats
+++ b/tests/smoke-test-general.bats
@@ -356,7 +356,7 @@ services:
 	#[[ $SKIP == 1 ]] && skip
 
         # Send all mail to /bin/true
-        echo "sendmail_path=/bin/true" >> .docksal/etc/php.ini
+        echo "sendmail_path=/bin/true" >> .docksal/etc/php/php.ini
         # Initialize the Project
         fin init
 	# Run fin share in a emulated terminal
@@ -374,6 +374,11 @@ services:
 	[[ "${output}" =~ "My Drupal 8 Site" ]]
 	unset output
 
+        # Clean Up
 	# Clean up kill ngrok session
         screen -X -S testNgrok quit
+        # Kill Docker Container
+        docker rm -f drupal8_web_1_ngrok
+        # Destroy Project
+        fin rm -f
 }
--- a/tests/smoke-test-general.bats
+++ b/tests/smoke-test-general.bats
@@ -351,3 +351,27 @@ services:
 	[[ "${output}" =~ "The VIRTUAL_HOST has been modified from feaTures.Alpha-beta_zulu.docksal to features.alpha-beta-zulu.docksal to comply with browser standards." ]]
 	unset output
 }
+
+@test "fin share" {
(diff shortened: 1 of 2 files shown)


tests/smoke-test-general.bats
```diff
@@ -350,4 +350,35 @@ services:
...
+        # Initialize the Project
+        fin init
+	# Run fin share in a emulated terminal
+	screen -S testNgrok -d -m fin share
+        # sleep so ngrok can load
+        sleep 10
+	# Query API for information
+	API=$(docker exec -it "drupal8_web_1_ngrok" sh -c "wget -qO- http://localhost:4040/api/tunnels")
+	# Return Public URL for site.
+	PUBLIC_HTTP_URL=$(echo "${API}" |  python -c 'import json,sys;obj=json.load(sys.stdin);print obj["tunnels"][0]["public_url"]')
```
lmakarov: Is this a Python json parser? If so, I suggest we use (and standardize on) [jq](https://stedolan.github.io/jq/) across the board. There are plans to include both jq and yq in `docksal/cli`

tests/smoke-test-general.bats
```diff
@@ -350,4 +350,35 @@ services:
...
\ No newline at end of file
+}
+
+@test "fin share" {
+	#[[ $SKIP == 1 ]] && skip
+
+        # Send all mail to /bin/true
+        echo "sendmail_path=/bin/true" >> .docksal/etc/php/php.ini
+        # Initialize the Project
+        fin init
```
lmakarov: Let's move this up, below `fin init` or `fin restart tests`, this way the extra init won't be necessary. Is does introduce a dependency on the previous test though, but we do have other tests happening linearly now, which does save time vs starting every test from scratch.

tests/smoke-test-general.bats
sean-e-dietrich: @lmakarov until we install jq within the container I'm using the standard Python json parser. I agree that if we can in fact get jq and yq installed within container then we can swap these out and may even want to make note of it for future. The problem is that ngrok previously broke and while 100% test coverage seems far away at the moment. I want to get 1 test in for `fin share` right now. Seeing as there have been individuals who made a note about it and more importantly because of @achekulaev recent blog post about `fin share`

tests/smoke-test-general.bats
sean-e-dietrich: @lmakarov I don't think we should introduce a dependency at all which is why I'm suggesting the `fin init` here to completely start from scratch on the project. That way regardless where in the line we run it. We know it will work.

tests/smoke-test-general.bats
lmakarov: `jq` is already there in Travis.

```
echo "${API}" | jq -r '.tunnels[0].public_url'
https://7c1b4c0a.ngrok.io
```

sean-e-dietrich: completely misunderstood and forgot the jq was being ran in travis and not the container. made the change

PR #681, opened 2018-07-03
Fixed tests for smoke general

sean-e-dietrich:
* Uncommented skip in `fin share` test
* Adjusted namespacing command test

### DIFF
--- a/tests/smoke-test-general.bats
+++ b/tests/smoke-test-general.bats
@@ -344,7 +344,7 @@ services:
 	echo "VIRTUAL_HOST=feaTures.Alpha-beta_zulu.docksal" > .docksal/docksal-local.env
 
 	# Check config (check if local environment variables are used in docksal.yml)
-	run fin config
+	run TERM=dumb fin config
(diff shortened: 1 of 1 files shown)


tests/smoke-test-general.bats
```diff
@@ -344,15 +344,14 @@ services:
 	echo "VIRTUAL_HOST=feaTures.Alpha-beta_zulu.docksal" > .docksal/docksal-local.env
 
 	# Check config (check if local environment variables are used in docksal.yml)
-	run fin config
-	echo "$output" | egrep "VIRTUAL_HOST: features.alpha-beta-zulu.docksal"
-	[[ $(echo "$output" | grep -c "feaTures.Alpha-beta_zulu.docksal") -eq 0 ]]
+	run TERM=dumb fin config
```
achekulaev: I guess it would be `TERM=dumb run fin config`

sean-e-dietrich: that ended up fixing it. thanks @achekulaev

lmakarov:
Thanks for fixing these @sean-e-dietrich

PR #682, opened 2018-07-04
Added docksal_ssh_agent to volumes for run-cli

sean-e-dietrich:
Currently when running `fin run-cli` it doesn't mount the ssh-agent therefore if you have ssh keys that are used for private git repos then they aren't included and you are prompted with a password for 

```
Enter passphrase for key '/home/docker/.ssh/id_rsa':
```

This is important if we want to try and convince people to use `run-cli` for things like using `terminus` or `npm` other mechanisms that require public key authentication.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -4515,6 +4515,7 @@ run_cli ()
 		${winpty} docker run --rm -it \
 			--mount type=bind,src=$(pwd),dst=/var/www \
 			--mount ${MOUNT_HOME} \
+			-v docksal_ssh_agent:/.ssh-agent:ro \
 			-e SECRET_SSH_PRIVATE_KEY \
 			-e SECRET_ACAPI_EMAIL \
 			-e SECRET_ACAPI_KEY \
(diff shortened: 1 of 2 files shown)


bin/fin
```diff
@@ -4515,6 +4515,7 @@ run_cli ()
 		${winpty} docker run --rm -it \
 			--mount type=bind,src=$(pwd),dst=/var/www \
 			--mount ${MOUNT_HOME} \
+			-v docksal_ssh_agent:/.ssh-agent:ro \
```
lmakarov: Please use `--mount` instead like the other mounts do.

lmakarov:
Good one. I'm actually surprised we did not have it there for so long.

lmakarov:
We should probably also add a test for this.

sean-e-dietrich:
would tend to agree, does the docksal org on github have private repositories?

lmakarov:
@sean-e-dietrich there is using real keys for tests (if that's what you though of).
Take a look at the ssh-agent tests here: https://github.com/docksal/docksal/blob/develop/tests/smoke-test-system.bats

sean-e-dietrich:
Thanks @lmakarov that is one way around it. I was thinking of a real world example where a git repo is downloaded but this makes sense as in order for the git repo to download the ssh-key has to be loaded.

PR #691, opened 2018-07-07
Document uuid-runtime dependency for Linux systems

valthebald:
Otherwise, fin throws an error uuidgen: command not found


sean-e-dietrich:
@valthebald we are actually getting rid of the uuid dependency in favor of a standard approach supported cross platforms.

https://github.com/docksal/docksal/blob/develop/bin/fin#L324

PR #695, opened 2018-07-10
Named db_data volume for MySQL/PostreSQL

lmakarov:
This gives database data volumes a name - `db_data`. This way they are easy to identify in the list of volumes (docker volume ls). E.g., `drupal8_db_data` vs `dd035157ff4f8ab90e60ece910c335c561caae6325382dc31435de56c0380346`

### DIFF
--- a/stacks/services.yml
+++ b/stacks/services.yml
@@ -32,6 +32,7 @@ services:
       - "${MYSQL_PORT_MAPPING:-3306}"
     volumes:
       - project_root:/var/www:ro,nocopy  # Project root volume (read-only)
+      - db_data:/var/lib/mysql  # Database data volume
     environment:
       - MYSQL_ROOT_PASSWORD=${MYSQL_ROOT_PASSWORD:-root}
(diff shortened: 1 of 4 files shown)


lmakarov:
I've just realized this may be a destructive change for existing projects. What if the new named volume replaced the existing anonymous one and all the data is gone. That would be an unpleasant surprise upon the update for most of the users.

Well, it looks like Docker Compose is actually smart about volume changes and won't destroy existing anonymous volumes:

```
$ DOCKSAL_VERSION=develop fin update
...
[STEP 6/6] Restarting running projects...
Starting services...
Creating volume "drupal8_db_data" with default driver
Recreating drupal8_db_1 ...
Recreating drupal8_cli_1 ... done
Recreating drupal8_web_1 ... done
Recreating drupal8_db_1  ... done
"db_data" has no effect. Remove the existing containers (with `docker-compose rm db`) to use the host volumemapping.
Waiting for drupal8_cli_1 to become ready...
Virtual Host: http://drupal8
Update finished
```

Phew... :)

PR #703, opened 2018-07-12
[WIP] New built-in `init` (dont merge)

achekulaev:
- Create new built-in init
- It cat be overridden


tests/smoke-test-general.bats
```diff
@@ -48,7 +48,28 @@ teardown() {
 	unset output
 }
 
-@test "fin init" {
+@test "fin init (built-in)" {
+	## [[ $SKIP == 1 ]] && skip
```
sean-e-dietrich: Should this be uncommented?

PR #706, opened 2018-07-13
New built-in `init`

achekulaev:
- Create new built-in init
- It can be overridden

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -5905,12 +5905,12 @@ addon_hook ()
 # $1 - name
 addon_get_path ()
 {
-	addon="$(get_addon_script $1)"
+	local addon="$(get_addon_script $1)"
 	if [[ ! -f "$addon" ]]; then
 		# If no addon was found then try command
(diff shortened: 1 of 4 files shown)


bin/fin
```diff
@@ -919,7 +938,7 @@ check_project_unique ()
 _confirm ()
 {
 	# Skip checks if not running interactively (not a tty or not on Windows)
-	if ! is_tty; then return 0; fi
+	if ! is_tty || is_travis; then return 0; fi
```
lmakarov: I think we should add a global `-y` flag to skip confirmations in `_confirm`.

Travis is not the only CI system that's using/faking interactive sessions and others may run into the need to skip confirmations.

Alternatively, commands can be run in a sub-shell in Travis to force a non-interactive environment, e.g. `run (fin init)` (I have not tested this, though).

achekulaev: Good notice. I'd like to keep "supporting all possible generic CIs" separately though. https://github.com/docksal/docksal/issues/707

PR #715, opened 2018-07-18
New exec_target = run-cli target for custom commands and addons

achekulaev:


docs/fin/custom-commands.md
```diff
@@ -164,6 +164,18 @@ services:
       - VIRTUAL_HOST
 ```
 
+## Executing commands inside standalone ad-hoc cli container
+
+This allows you to define a global command, that will execute in a standalone cli container spawned on demand.
```
shelane: This comma isn't needed here.

PR #748, opened 2018-08-01
Improve exec, make it more robust

### LINKED ISSUE #738: Arguments passed to `fin exec` are inconsistent with `fin run-cli`
**Description**
Running Docksal 1.60.2 and docksal/cli:2.3-php7.2.

Using `fin run-cli [args]` works as you expect: a standalone cli container is spawned and `[args]` is fed to it as if you ran `fin bash [new-container]` and entered the command manually.

However, `fin exec [args]` is problematic in that it has its own logic:

If argument `$1` is an existing file, it is passed to `bash -c` _without_ any of the other arguments passed to it. Example:

`fin exec path/to/script.sh arg1 arg2`

Those `arg1` and `arg2` arguments are never passed to `script.sh`. Additionally, if you try to pass a single argument to be executed as a shell statement, it fails because it escapes the entire argument:

`fin exec "source ~/.profile && npm run" something`

The fact that `something` is outside of the quotes is caused by Bash pulling it to the side when using `$@` in a custom command, for example:

`fin exec "source ~/.profile && npm ${@}"`

**Suggested solution**
In my opinion, `fin exec [args]` should behave exactly like `fin bash` followed by entering `[args]` by hand and pressing [enter]. I think this is what people would expect from such a command.

In order to support this, the only thing `fin exec [args]` should do is run a `docker exec` in the right container with `bash -c "[args]"` where `"[args]"` is one big string containing all the remaining arguments passed to `fin exec`. No _additional_ escaping should take place and there should not be a check if a file exists and then run that in a different flow.

My guess is that the `fin exec [file]` format was introduced to help people running scripts that do not have the executable bit set, but it might help to just err out with a message asking the user to set that bit.

achekulaev:
Closes #738

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -4454,7 +4454,7 @@ _exec ()
 	# ------------------------------------------------ #
 	# Convert array of parameters into escaped string
 	local cmd="$1"
-	if [[ -n $2 ]]; then
+	if [[ $# > 1 ]]; then
 		args=$(printf "%q " "${@:2}")
 	fi
 	# ------------------------------------------------ #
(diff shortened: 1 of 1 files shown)


achekulaev:
Closing without merging. See #738 for explanations.

Issue #210, opened 2017-05-06
Tests should be updated


achekulaev:
* `fin project ...`
* `fin db ...` (remove `fin sql...`)
* `fin hosts`
* `fin addon install/remove example`
* `fin system`
* `fin image`
* listing of custom commands and addons in `fin help`
* parse_params

achekulaev:
* `fin system`
* `fin config get`
* `fin config set`
* `fin config remove`

lmakarov:
@achekulaev please review and split as discussed.

lmakarov:
RIP

achekulaev:
Did you split it?

lmakarov:
> Did you split it?

No. We've been pushing this out for so long, that it does not make sense to keep it around.
Smaller tickets would make sense so long as we plan on tackling them soon and have particular assignees (e.g., based on who created/maintains a particular feature in fin).

PR #750, opened 2018-08-01
Fix help for project/global commands on Ubuntu 18.04.

vaza18:
`fin help` console output
Before:
image

After:
image


lmakarov:
@vaza18 thanks for the PR!

We should probably add a test case for this.

achekulaev:
@lmakarov I added it to the lists of tests into #210 to tackle during 1.11.0 development.

Issue #762, opened 2018-08-02
Virtual hosts comply with browser standards in a multi-site multi-domain setup


maikel-finck:
**Description**

There isn't a cooked solution at the docs on how to use a multi-site multi-domain environment with a single code base. A quick workaround was to declare the domains in the virtual_host variable. This may not be an "used as intended" scenario but is one that works with ACSF.

After this pull request this is not working anymore, it seems that virtual_host now expect only a single domain, is that correct?
https://github.com/docksal/docksal/pull/653/files/ae1e447ef25de6f29ea0013fe8cb11e3881ef837

**Steps to reproduce the issue:**

1. Create a new environment, extend Acquia stack.
2. Set up VIRTUAL_HOST with multiple domains, comma separated.
Concept example:
VIRTUAL_HOST="site1.domain1,site2.domain1,site2.domain2,siteN.domainN,tool1.tools,tool2.tools"

3. Override your YML service labels and distribute the domains
Concept example:
service: web
io.docksal.virtual-host="site1.domain1,site2.domain1,site2.domain2,siteN.domainN"
service: tool1
io.docksal.virtual-host="tool1.tools"
service: tool1
io.docksal.virtual-host="tool1.tools"

4. Start your environment

**Describe the results you received:**

WARNING:  The VIRTUAL_HOST has been modified from
site1.domain1,site2.domain1,site2.domain2,siteN.domainN,tool1.tools,tool2.tools
to
site1.domain1site2.domain1site2.domain2siteN.domainNtool1.toolstool2.tools
to comply with browser standards.

**Describe the results you expected:**

The multiple domains would still work as expected before this fix was published.

<br>

**Output of `fin sysinfo`:**

<details>
  <summary>fin sysinfo output</summary>

  ```
WARNING: No swap limit support
███  OS
Linux Ubuntu 16.04
Linux lnxcit011459 4.8.0-41-generic #44~16.04.1-Ubuntu SMP Fri Mar 3 17:11:16 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux

███  ENVIRONMENT
MODE : Linux Kernel
DOCKER_HOST : 

███  FIN
fin version: 1.67.2

███  DOCKER COMPOSE
EXPECTED VERSION: 1.21.1
docker-compose version 1.21.1, build 5a3f1a3
docker-py version: 3.3.0
CPython version: 3.6.5
OpenSSL version: OpenSSL 1.0.1t  3 May 2016

███  DOCKER
[...]

PR #763, opened 2018-08-02
Added test to confirm virtual_host can accept multiple hosts names

sean-e-dietrich:
Fixes issue: https://github.com/docksal/docksal/issues/762

When multiple hostnames are defined in `VIRTUAL_HOST`


lmakarov:
Multiple hostnames should not be set via `VIRTUAL_HOST `. We'll should update the docs to state this.

PR #767, opened 2018-08-03
Add redis service.

michael-molchanov:


achekulaev:
LGTM

sean-e-dietrich:
This looks good. I'd like to see if there is more of an ability to configure redis a little better. Using the standard redis image right now basically leaves no room for project configuration so for projects where they need to tune redis they can't do it.

achekulaev:
Good point @sean-e-dietrich

lmakarov:
Wait... We already have docs for using Redis here: https://docs.docksal.io/en/master/tools/redis/
We just need to choose which image to use as the default in the services.yml and update the docs respectively.

achekulaev:
Closing in favor of https://github.com/docksal/docksal/issues/874

PR #780, opened 2018-08-18
Exit if system services failed to start

### LINKED ISSUE #772: Failure to pull system image does not result in failure of Docksal installation
**Description**

We use a custom `IMAGE_DNS` for Docksal which is pulled during the regular Docksal installation. We noticed that if docker is unable to pull this image, Docksal still continues as if nothing bad happened and does not give an exit status that indicates that something failed.

1. Set `IMAGE_DNS` to an image on a Docker registry that is unreachable
2. Install Docksal using `curl -fsSL get.docksal.io | bash`

The image pull fails and Docksal returns exit status 0.

Docksal installation fails and provides an exit status != 0.

GMTA:
This prevents broken system states where one or more Docksal services failed to
start without the user being informed of this being the case.

Examples of causes:
- Network issues
- Providing the wrong image name for the system services

Fixes #772.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -3296,6 +3296,7 @@ install_proxy_service ()
 		--mount type=bind,src=/var/run/docker.sock,dst=/var/run/docker.sock \
 		${mount_certs} \
 		"${IMAGE_VHOST_PROXY}" >/dev/null
+	if_failed_error "Failed starting the proxy service."
 }
 
 # Detects upstream DNS or falls back to DOCKSAL_DEFAULT_DNS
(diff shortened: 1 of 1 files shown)


achekulaev:
SSH Agent service is not so critical to prevent the whole system from working in case it fails to start

GMTA:
@achekulaev So DNS and Proxy are pretty important. At least there should be a clear distinctive message when SSH fails, right?

achekulaev:
As far as I know Docker will post a pretty clear and distinctive message if it fails to tart a container, no?

lmakarov:
`docksal-vhost-proxy` is vital, since pretty much nothing will work without it
`docksal-dns` is important, but not ciritical
`docksal-ssh-agent` - may be critical for some setups and nice-to-have for others

So are already doing a soft-fail handling and restarting these automatically if they are not running:
https://github.com/docksal/docksal/blob/d3b719a02950d4bd2e2eadde77da7bb97c095939/bin/fin#L824-L832

Should we turn this into a hard fail instead? Most likely, yes. @achekulaev - what's your opinion?

GMTA:
@achekulaev

> As far as I know Docker will post a pretty clear and distinctive message if it fails to tart a container, no?

Well, not really (it is just as distinctive as all the other lines scrolling along). I would be okay-ish with making it a clear distinctive red color without actually stopping the process.

But my main gripe and reason for this PR is that the process returns exit status 0 although the services were not started correctly. An alternative solution to this would be to introduce an option like `--require-services-started` (or env var) that would stop the process if something went wrong. Use case: we automate the installation of Docksal and would like to make sure that the images Docksal uses are pulled correctly and the containers are running. Workaround for this is to manually check if the containers are running.

PR #781, opened 2018-08-20
Updated Linux requirement for non-persistent network configuration

frederickjh:
This should have been noted in the change log too.


lmakarov:
@frederickjh `netplan` is a Ubuntu-only tool. Docksal does not require or use it.

PR #782, opened 2018-08-20
Fix detecting first nameserver in resolv.conf

GMTA:
The comments said that the first suitable nameserver is used, but it actually took the _last_ suitable nameserver in `resolv.conf`. This PR fixes that behavior.


achekulaev:
OMG thanks!!

PR #783, opened 2018-08-22
Fix wrong name of some blackfire env variables

ApacheEx:
Hola 👋

if we check https://github.com/docksal/docksal/blob/develop/stacks/services.yml

```yml
 # Blackfire
  blackfire:
    image: blackfire/blackfire
    environment:
      - BLACKFIRE_SERVER_ID
      - BLACKFIRE_SERVER_TOKEN
```

we can notice that variables are 
`BLACKFIRE_SERVER_ID` and `BLACKFIRE_SERVER_TOKEN` instead of their `SECRET_` version.

so, there are 2 options:
1. fix docs (**done** in this PR)
2. fix services.yml

### DIFF
--- a/docs/tools/blackfire.md
+++ b/docs/tools/blackfire.md
@@ -15,8 +15,8 @@ With this option, the API keys are stored on your host and not exposed in the pr
 ```bash
 fin config set --global SECRET_BLACKFIRE_CLIENT_ID=<blackfire-client-id>
 fin config set --global SECRET_BLACKFIRE_CLIENT_TOKEN=<blackfire-client-token>
-fin config set --global SECRET_BLACKFIRE_SERVER_ID=<blackfire-server-id>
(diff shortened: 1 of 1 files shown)


lmakarov:
Quick and dirty answer - yes, the docs should be fixed.

Long term answer - it's complicated. We introduced the notion of secrets recently (variables prefixed with `SECRET_`) to protect API keys in stack configuration showing up in plain text. `fin` hides them in the `fin config` output. At the same time, these variables need to be present in the containers using them without the `SECRET_` prefix. `docksal/cli` handles this at startup and re-exported secrets inside `cli` without the prefix.

We could do one of the following (long term):
- fork `blackfire/blackfire` and add the variable conversion logic there
- figure out another way of hiding secrets in the `fin config` output
- figure out another smart way of converting variables inside a container without forking it

achekulaev:
I guess for the time being while we figure out the way to go with the container, the docs should be updated.

Issue #787, opened 2018-09-12
Remove cli_home volume whenever cli image version is updated


lmakarov:
`cli_home` is an anonymous volume mounted in the `cli` service:

https://github.com/docksal/docksal/blob/a7fc9358c3e85e2bfc6e810a187c78c625f32c05/stacks/services.yml#L91

It was introduced to optimize write heavy operations in `/home/docker` inside `cli`. See https://github.com/docksal/docksal/issues/325

When a docker compose service configuration (including the image) is updated, volumes are not dropped and are remounted as-is. This is usually desirable, e.g., when updating the `db` service settings, the database(s) date volume is preserved.

In the case of `cli`, however, we do want the `cli_home` volume dropped. That is because this volume contains a bunch of tools and configs that tend to change from one `docksal/cli` version to another. Without dropping it, the update won't be complete and will miss the changes in `/home/docker`.

At the same time (and ideally) we don't won't to drop the volume on EVERY configuration update to the `cli` service. Only image version changes require `cli_home` to be dropped, which makes this somewhat complicated, since we currently do not have a reliable way to track stack configuration changes.

lmakarov:
We already instruct users in docs to do `fin project reset cli` when changing PHP versions:
https://docs.docksal.io/en/master/advanced/stack-settings/#php-versions

The issues is that this stopped working after `/home/docker` [became](https://github.com/docksal/docksal/commit/81a42f4da48ee2aacc37198103ac124be7964a4e) a named volume (`<project>_cli_home`). `docker-compose rm -vf <container>` will only drop anonymous volume and won't touch named once.

Looks like, we'll need a special handling for `fin project reset cli` and manually drop the `<project>_cli_home` volume.

achekulaev:
Not sure what activities left here but it is in the ToDo column for 1.12.

lmakarov:
This is still a valid issue when the docksal/cli version is updated in the default stacks (like with the 1.11 release, actually). When this happens, `<project>_cli_home` should be dropped, otherwise updated in `/home/docker` from the new `docksal/cli` version won't propagate.

PR #789, opened 2018-09-19
Drop volume when resetting db container

ToonSpinISAAC:
This PR changes the behavior of the command `fin reset db`. The usage string in the command line says `fin reset` does this:

    reset [service]            Recreate all or specified project services, their containers and named volumes
                               Changes to home directory in `cli` are preserved.

So I expect `fin reset db` to throw away the database volumes. Instead, what it currently does is keep the volume so the only easy way to get rid of them is to reset your entire project. The reason it works this way this is that it used `docker-compose rm -v` to remove the volumes, but the DB volume is not anonymous.

I've interpreted this to be a bug and fixed it in this pull request. In this PR, `fin reset db` does what it previously did, and also removes the database volume.

If there are any questions or remarks, please go ahead! :)

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1475,8 +1475,7 @@ show_help_project() {
 	echo
 	printh "status" "List project services (alias: fin ps)"
 	printh "restart" "Restart project services (alias: fin restart)"
-	printh "reset [service]" "Recreate all or specified project services, their containers and named volumes"
-	printh "" "Changes to home directory in \`cli\` are preserved."
+	printh "reset [service]" "Recreate all or specified project services, their containers and volumes"
 	echo
 	printh "remove [option] [service]" "Remove all project services, networks and all their volumes, or specified services only"
 	printh "    rm [option] [service]" ""
--- a/bin/fin
+++ b/bin/fin
@@ -1055,10 +1055,11 @@ _remove_containers ()
 		# Removing requested containers only
 		docker-compose kill "$@" &&
 			docker-compose rm -vf "$@" &&
-			# For cli, we have to manually drop the home volume, since named volumes are omitted by the "-v" option below
+			# For cli and db, we have to manually drop the volumes, since named volumes are omitted by the "-v" option above
 			{
 				for i in "$@"; do
 					[[ "$i" == "cli" ]] && docker volume rm -f "${COMPOSE_PROJECT_NAME_SAFE}_cli_home" >/dev/null
+					[[ "$i" == "db" ]] && docker volume rm -f "${COMPOSE_PROJECT_NAME_SAFE}_db_data" >/dev/null
 				done
 			}
 	fi


lmakarov:
Yep, makes total sense.

Also, the statement about "Changes to home directory in `cli` are preserved." will need to be dropped.
We ask users to reset `cli` when switching PHP/image versions and that does require the `<project>_cli_home` volume (`/home/docker` directory inside cli) dropped. @achekulaev any thoughts or objections?

ToonSpinISAAC:
@lmakarov thanks for looking! Would you like me to drop that phrase in this PR?

lmakarov:
@ToonSpinISAAC yes, let's change it to just "Recreate all or specified project services, their containers and volumes"

ToonSpinISAAC:
Okay, I've just pushed a commit.

achekulaev:
@lmakarov 

```
We ask users to reset cli when switching PHP/image versions and that does require the <project>_cli_home volume (/home/docker directory inside cli) dropped
```

No it does not require home volume dropped.
And the volume is not dropped currently. Proof:

```
15:41:09 ~/Projects/d8test
§ fin exec touch /home/docker/newfile

15:41:24 ~/Projects/d8test
§ fin reset cli
Killing d8test_cli_1 ... done
Going to remove d8test_cli_1
Removing d8test_cli_1 ... done
Starting services...
d8test_db_1 is up-to-date
Creating d8test_cli_1 ... done
Recreating d8test_web_1 ... done
Waiting for d8test_cli_1 to become ready...

15:41:45 ~/Projects/d8test
§ fin exec ls -la /home/docker/newfile
-rw-r--r-- 1 docker dialout 0 Sep 20 19:41 /home/docker/newfile

```

So I do not see a reason to drop the wording.

bin/fin
```diff
@@ -1474,8 +1475,7 @@ show_help_project() {
 	echo
 	printh "status" "List project services (alias: fin ps)"
 	printh "restart" "Restart project services (alias: fin restart)"
-	printh "reset [service]" "Recreate all or specified project services, their containers and named volumes"
-	printh "" "Changes to home directory in \`cli\` are preserved."
+	printh "reset [service]" "Recreate all or specified project services, their containers and volumes"
```
achekulaev: The wording about "Changes to home directory" needs to be returned. Please see conversation for details.

lmakarov:
@achekulaev the `cli_home` volume has to be dropped (see #787).
That has been implemented in aed2a3c979e8c1d3a9aeb8531659db70ce04990b

achekulaev:
Ok I see. I guess the change related to the `cli` still should not be part of this PR. This PR is DB related and we're mixing stuff up.
I am not too hard on this point though.

lmakarov:
Agreed, but I think it's fine as is, so we don't go back and forth on this. Let's just merge.

PR #808, opened 2018-11-01
Docs migration to Hugo and Netlify

lmakarov:
- Using `hugo` instead of `mkdocs` as a static site generator
- Using Netlify instead of ReadTheDocs for hosting
- Restructured docs sections

Preview: https://feature-hugo-docs--focused-shannon-c08e59.netlify.com


lmakarov:
Moved Algolia search integration into a separate issue: https://github.com/docksal/docksal/issues/809

PR #815, opened 2018-11-08
Fixed issue related to RC check

sean-e-dietrich:
Uses the tag version to grab files if `DOCKSAL_USE_RC` is set.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -4269,6 +4269,7 @@ update ()
 		echo-green "[STEP 1/$TOTAL_STEPS] Updating fin..."
 
 		local fin_url=$(get_fin_url)
+		wants_rc_version && DOCKSAL_VERSION=$(get_docksal_latest_version)
 		local new_fin
 		new_fin=$(curl -kfsSL "${fin_url}?r=$RANDOM")
 		if_failed_error "fin download failed."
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -4372,13 +4373,18 @@ update ()
...
 
-# Return latest fin download url for the latest tag
-get_latest_fin_url ()
+# Return the latest version
+get_docksal_latest_version ()
 {
 	# Get the latest release tag
 	URL_REPO_RELEASE="${GITHUB_API}/repos/docksal/docksal/releases"
-	local latest_tag=$(curl -sSL $URL_REPO_RELEASE | grep tag_name | sed 's/\"tag_name": \"\(.*\)",/\1/' | head -n 1 | tr -d '[:space:]')
+	echo $(curl -sSL $URL_REPO_RELEASE | grep tag_name | sed 's/\"tag_name": \"\(.*\)",/\1/' | head -n 1 | tr -d '[:space:]')
```
sean-e-dietrich: @achekulaev @lmakarov Does it make more sense to set `DOCKSAL_VERSION` here instead of doing another API call here https://github.com/docksal/docksal/pull/815/files#diff-41adca34c3eac75ca9453c7cb0634f6aR4272

PR #820, opened 2018-11-13
Update VirtualBox requirement to 5.2.20 for macOS Mojave compatibility

achekulaev:
- boot2docker 18.09 has guest tools for 5.2.20

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -115,8 +115,8 @@ REQUIREMENTS_WINPTY='0.4.3'
 REQUIREMENTS_WINPTY_CYGWIN='2.8.0'
 
 # VirtualBox download URLs
-URL_VBOX_MAC="http://download.virtualbox.org/virtualbox/5.2.2/VirtualBox-5.2.2-119230-OSX.dmg"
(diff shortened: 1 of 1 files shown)


lmakarov:
There is also a new release of Docker Machine:
https://github.com/docker/machine/releases/tag/v0.16.0

Should be a safe update.

achekulaev:
@lmakarov I am not open to updating so many tools. We were supposed to deploy a release tomorrow. Updating now doesn't give us time to even test it ourselves, and I postponing the release yet again does not sound like a plan to me. 

I would like to release. The only reason I'm updating VBOx now is because it is a blocking issue, and because I feel pretty safe about updating VirtualBox, but updating any Docker related tools is a risky endeavor, and example of docker-compose highlights it as well as possible.

lmakarov:
@achekulaev ok, I did not realize you were planning this for the 1.11 release, since you mentioned boot2docker 18.09. We will have a mismatch in the VBox guest additions, since boot2docker 18.06.1 uses v5.2.2. However, Mojave users have to install VBox v5.2.20 anyway to get it to work. So yes, let's have this merged as is.

Issue #817, opened 2018-11-10
Running fin restart service disconnects all containers


sean-e-dietrich:
When trying to restart one service example `fin restart cli` or `fin restart web` it disconnects all of the containers from the network. So you actually need to run a run a full `fin restart` on the project.

```
➜  golden_gate_national_parks_conservancy git:(master) ✗ fin start
Starting services...
Starting golden_gate_national_parks_conservancy_redis_1 ... done
Starting golden_gate_national_parks_conservancy_cli_1   ... done
Starting golden_gate_national_parks_conservancy_db_1    ... done
Starting golden_gate_national_parks_conservancy_web_1   ... done
Waiting for golden_gate_national_parks_conservancy_cli_1 to become ready...
Virtual Host: http://ggnpc.docksal

➜  golden_gate_national_parks_conservancy git:(master) ✗ fin status
                     Name                                   Command                  State                 Ports
---------------------------------------------------------------------------------------------------------------------------
golden_gate_national_parks_conservancy_cli_1     /opt/startup.sh supervisord      Up (healthy)   22/tcp, 3000/tcp, 9000/tcp
golden_gate_national_parks_conservancy_db_1      /entrypoint.sh mysqld            Up             0.0.0.0:32768->3306/tcp
golden_gate_national_parks_conservancy_redis_1   /docker-entrypoint.sh redi ...   Up             6379/tcp
golden_gate_national_parks_conservancy_web_1     httpd-foreground                 Up             443/tcp, 80/tcp

➜  golden_gate_national_parks_conservancy git:(master) ✗ fin restart web
Disconnecting project network...
Stopping services...
Stopping golden_gate_national_parks_conservancy_web_1 ... done
Starting services...
Creating network "golden_gate_national_parks_conservancy_default" with the default driver
golden_gate_national_parks_conservancy_db_1 is up-to-date
golden_gate_national_parks_conservancy_redis_1 is up-to-date
golden_gate_national_parks_conservancy_cli_1 is up-to-date
Starting golden_gate_national_parks_conservancy_web_1 ... done
Connected vhost-proxy to "golden_gate_national_parks_conservancy_default" network.

➜  golden_gate_national_parks_conservancy git:(master) ✗ fin status
                     Name                                   Command                  State            Ports
----------------------------------------------------------------------------------------------------------------
golden_gate_national_parks_conservancy_cli_1     /opt/startup.sh supervisord      Up (healthy)
[...]

PR #823, opened 2018-11-14
Use "docker-compose restart" in "fin project restart"

lmakarov:
Partially solves #817


achekulaev:
We did stop/start for a reason. I cannot immediately recall a reason, but stop/start is very different from restart. If I remember correctly reloading config is one of the reasons.

PR #843, opened 2018-11-19
Support for docker-compose 1.23 new naming schema

achekulaev:


achekulaev:
Composer version change with tests will be in a separate PR

PR #853, opened 2018-11-22
Upgrade docker-compose to 1.23

### LINKED ISSUE #819: Update to Docker Compose 1.23.0
**Update:**

Docker Compose v1.23.0 has changed it's naming scheme which breaks existing assumptions about container names in a Docker Compose managed stack (reported in docker/compose#6316)

---

Database commands rely on the container name being `${COMPOSE_PROJECT_NAME_SAFE}_db_1`

There have been times where docker compose will append a hash to the end of the container ID and therefore those containers are no longer usable with the `fin db ...` commands.

The following was a conversation in Drupal Slack 

```
fabean [12:38 PM]
@sean.e.dietrich different weird issue. Now (I assume unrelated) all my containers are `projectname_cli_1-HASH` which is breaking `fin` commands from working with those containers.

2018-11-13-144005_grim.png 

This is on linux. I tried `fin system stop` then `docker system prune` I've restarted docker, and my computer. I also tried spinning up a new site and get the same sort of things on the end of my container names

sean.e.dietrich [12:39 PM]
can you run `fin --version`

fabean [12:40 PM]
1.69.1

sean.e.dietrich [12:40 PM]
what happens when you run `fin restart`

fabean [12:41 PM]
@sean.e.dietrich it restarts those project containers with the hashes still on the end

sean.e.dietrich [12:42 PM]
you mentioned that `fin` is not working with the hashes correct at the end of the containers?

fabean [12:42 PM]
yes, for example running `fin db create databasename`
says `Error: No such container: scce-drupal_db_1`

sean.e.dietrich [12:44 PM]
ahh ok
are you at a point you can destroy the db container and rebuild it?

fabean [12:45 PM]
Yes I can delete everything no problem

sean.e.dietrich [12:45 PM]
if you run `fin reset db` that should rebuild the database container

fabean [12:48 PM]
Yes it does, same issue of having hashes on the end of my containers though.
```

2018-11-13-144005_grim

achekulaev:
Closes #819


lmakarov:
@achekulaev let's upgrade to the latest version - 1.23.1

PR #868, opened 2018-12-01
Updated README and Makefile

sean-e-dietrich:
Included information for people who are using boot2docker (VirtualBox).

Was having issues learning how to build the docs locally when I found the Make file references docker directly. I am working with Boot2Docker. Made all of the commands set to defaults if not defined. Additionally, added info on committing updates to the docs.

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,7 +1,7 @@
-DOCKER       = docker
-HUGO_VERSION = 0.50
-DOCKER_IMAGE = jojomi/hugo:$(HUGO_VERSION)
-DOCKER_RUN   = $(DOCKER) run --rm --interactive --tty --volume $(PWD):/src
+DOCKER       ?= docker
+HUGO_VERSION ?= latest
+DOCKER_IMAGE ?= jojomi/hugo:$(HUGO_VERSION)
+DOCKER_RUN   ?= $(DOCKER) run --rm --interactive --tty --volume $(PWD):/src
 
 .PHONY: all build build-preview help serve test
 
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -1,5 +1,5 @@
 DOCKER       ?= docker
-HUGO_VERSION ?= latest
+HUGO_VERSION ?= 0.50
 DOCKER_IMAGE ?= jojomi/hugo:$(HUGO_VERSION)
 DOCKER_RUN   ?= $(DOCKER) run --rm --interactive --tty --volume $(PWD):/src
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -20,13 +20,13 @@ For more info, see [Directory Structure](https://gohugo.io/getting-started/direc
 
 ## Building
 
-Before starting you must install the [Learn Theme](https://github.com/matcornic/hugo-theme-learn). This can be done by running `git submodule update` within the local repository.
+Before starting, you must install the [Learn Theme](https://github.com/matcornic/hugo-theme-learn). This can be done by running `git submodule update` within the local repository.
 
 Building and previewing docs is done via a docker container.
 
 Run `make serve` and point your browser to http://localhost:1313 to open the site.
 
-If you do not have Docker installed natively and are using VirtualBox you can run using the `fin` commmand. When accessing you would use the IP address assigned to your VirtualBox machine (ie. http://192.168.64.100:1313).
+If you do not have Docker installed natively and are using VirtualBox, you can run using the `fin` commmand. When accessing, you would use the IP address assigned to your VirtualBox machine (i.e., http://192.168.64.100:1313).
 
 ```bash
 DOCKER="fin docker" make serve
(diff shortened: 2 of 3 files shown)


docs/Makefile
```diff
@@ -1,7 +1,7 @@
-DOCKER       = docker
-HUGO_VERSION = 0.50
-DOCKER_IMAGE = jojomi/hugo:$(HUGO_VERSION)
-DOCKER_RUN   = $(DOCKER) run --rm --interactive --tty --volume $(PWD):/src
+DOCKER       ?= docker
+HUGO_VERSION ?= latest
```
lmakarov: Two issues with using `latest`:

1.  Things may break with a new version of Hugo
2. `latest` tag may be different for different users, unless everyone always does a `docker pull`

docs/README.md
```diff
@@ -20,10 +20,18 @@ For more info, see [Directory Structure](https://gohugo.io/getting-started/direc
 
 ## Building
 
+Before starting, you must install the [Learn Theme](https://github.com/matcornic/hugo-theme-learn). This can be done by running `git submodule update` within the local repository.
```
lmakarov: The theme is linked as a git submodule in `themes/learn` to https://github.com/docksal/hugo-theme-learn, which is a fork of https://github.com/matcornic/hugo-theme-learn.

If changes are needed to the theme, then they would happen in the theme repo first, then re-linked in this repo. This process is somewhat heavy, but it's the only one that would allow for easy updates from upstream. I'm open to suggestion to improve/simplify this flow. We should also add this to the docs here.

I this we should document the recommendation from [here](https://stackoverflow.com/questions/4611512/is-there-a-way-to-make-git-pull-automatically-update-submodules) to configure git to pull submodules automatically:

```
git config --global submodule.recurse true
```

docs/README.md
```diff
@@ -20,10 +20,18 @@ For more info, see [Directory Structure](https://gohugo.io/getting-started/direc
...
+
 Building and previewing docs is done via a docker container.
 
 Run `make serve` and point your browser to http://localhost:1313 to open the site.
 
+If you do not have Docker installed natively and are using VirtualBox, you can run using the `fin` commmand. When accessing, you would use the IP address assigned to your VirtualBox machine (i.e., http://192.168.64.100:1313).
+
+```bash
+DOCKER="fin docker" make serve
```
lmakarov: 👍

docs/Makefile
```diff
@@ -1,7 +1,7 @@
-DOCKER       = docker
-HUGO_VERSION = 0.50
-DOCKER_IMAGE = jojomi/hugo:$(HUGO_VERSION)
-DOCKER_RUN   = $(DOCKER) run --rm --interactive --tty --volume $(PWD):/src
+DOCKER       ?= docker
+HUGO_VERSION ?= 0.50
```
sean-e-dietrich: @lmakarov I changed it back to `0.50` but believe that we should make it overridable.

docs/README.md
```diff
@@ -20,10 +20,24 @@ For more info, see [Directory Structure](https://gohugo.io/getting-started/direc
...
+
+```
+git config --global submodule.recurse true
+```
+
 Building and previewing docs is done via a docker container.
 
 Run `make serve` and point your browser to http://localhost:1313 to open the site.
 
+If you do not have Docker installed natively and are using VirtualBox, you can run using the `fin` commmand. When accessing, you would use the IP address assigned to your VirtualBox machine (i.e., http://192.168.64.100:1313).
```
shelane: You can run? I'm not reading this sentence correctly or it's missing something.

sean-e-dietrich: Yes, that was pretty horrible grammar, thinking of changing to:

If you do not have Docker installed natively but are using VirtualBox for your Docksal projects, the `fin` commmand can be used. When accessing the built project, you would use the IP address assigned to your VirtualBox machine (i.e., http://192.168.64.100:1313).

shelane: So `fin` can be used with the make command?

sean-e-dietrich: @achekulaev wasn't thinking either of those. I was just suggesting running:

```
 DOCKER="fin docker" make serve
```

As I have `make` already just didn't have the docker command.

achekulaev: Okay now I understood the context. The sentence needs to be revised. "If you don't have Docker then fin command can be used" sounds like `fin` is somehow replacement for Docker.

I would suggest something like:

If you use Docksal via VirtualBox, then `fin docker` command can be used as a replacement for Docker binary e.g. ` DOCKER="fin docker" make serve`..."

sean-e-dietrich:
@lmakarov What do you think about merging this PR in?

docs/README.md
achekulaev: @sean-e-dietrich I still believe the sentence needs to be revised. I can do it for you if you don't mind.

sean-e-dietrich: @achekulaev I don't mind at all. It may be easier if you could please.

PR #889, opened 2018-12-10
Remove COLUMNS/LINES workaround

GMTA:
Originally introduced by 6e53837d3 and other commits; this workaround is no
longer necessary since the bug in Docker was squashed:

https://github.com/moby/moby/issues/35407

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -4624,14 +4624,13 @@ _exec ()
 
 	# Enter project containers
 	# Use the docker user in cli (-u docker) instead of root (default user).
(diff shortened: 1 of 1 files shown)


achekulaev:
We would need to thoroughly check this first

achekulaev:
I am inclined not to do this change basing on this comment: https://github.com/moby/moby/pull/37172#issuecomment-406844485

Looks like there are still cases when it is still beneficial to have those variables.

cc @lmakarov @sean-e-dietrich

PR #893, opened 2018-12-12
adds Algolia search to documentation

### LINKED ISSUE #809: Algolia search integration for new docs
https://www.algolia.com

shelane:
Closes #809

Additional css was added to keep documentation from bleeding through to the results box and keeping the size and padding of the search box as it was.

### DIFF
--- a/docs/layouts/partials/menu-footer.html
+++ b/docs/layouts/partials/menu-footer.html
@@ -15,6 +15,6 @@
  apiKey: '656714d8c3d5f7138b76633780faf6ba',
  indexName: 'docksal',
  inputSelector: '#search-by',
- debug: true // Set debug to true if you want to inspect the dropdown
+ debug: false // Set debug to true if you want to inspect the dropdown
  });
 </script>
--- a/docs/layouts/partials/menu-footer.html
+++ b/docs/layouts/partials/menu-footer.html
@@ -17,4 +17,15 @@
  inputSelector: '#search-by',
  debug: false // Set debug to true if you want to inspect the dropdown
  });
+ jQuery(document).ready(function() {
+     /*
+      * Prevent left and right arrow keys from engaging previous or next links.
+      */
+     jQuery('input').keydown(function (e) {
+         // left and right arrow keys
+         if (e.which == '37' || e.which == '39') {
+             e.stopPropagation();
+         }
+     });
+ });
 </script>
--- a/docs/layouts/partials/menu-footer.html
+++ b/docs/layouts/partials/menu-footer.html
@@ -9,3 +9,12 @@
 </div>
 <!-- Place this tag in your head or just before your close body tag. -->
 <script async defer src="https://buttons.github.io/buttons.js"></script>
+<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
+<script type="text/javascript">
+ docsearch({
+ apiKey: '656714d8c3d5f7138b76633780faf6ba',
+ indexName: 'docksal',
+ inputSelector: '#search-by',
+ debug: true // Set debug to true if you want to inspect the dropdown
+ });
+</script>
--- a/docs/layouts/partials/custom-header.html
+++ b/docs/layouts/partials/custom-header.html
@@ -1,1 +1,2 @@
 {{ partial "google-tag-manager.html" . }}
+<link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.css" />
--- a/docs/layouts/partials/search.html
+++ b/docs/layouts/partials/search.html
@@ -1,4 +1,4 @@
-<div class="searchbox">
+<div class="searchboxed">
     <label for="search-by"><i class="fas fa-search"></i></label>
     <input data-search-input id="search-by" type="text" placeholder="{{T "Search-placeholder"}}">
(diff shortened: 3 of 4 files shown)


achekulaev:
We'll need to wait for Leonid to come back from vacation and take a look.

sean-e-dietrich:
I just tested this out on the following URL https://deploy-preview-893--docs-docksal-io.netlify.com/

screen shot 2018-12-12 at 8 02 51 am

Seems like as you are typing the old search also runs with Algolia. Not sure if we can turn that off.

shelane:
Hmmm. I didn’t see that in my tests. We can hide it with css if we can’t turn the js off.

docs/layouts/partials/menu-footer.html
```diff
@@ -9,3 +9,12 @@
 </div>
 <!-- Place this tag in your head or just before your close body tag. -->
 <script async defer src="https://buttons.github.io/buttons.js"></script>
+<script type="text/javascript" src="https://cdn.jsdelivr.net/npm/docsearch.js@2/dist/cdn/docsearch.min.js"></script>
+<script type="text/javascript">
+ docsearch({
+ apiKey: '656714d8c3d5f7138b76633780faf6ba',
+ indexName: 'docksal',
+ inputSelector: '#search-by',
+ debug: true // Set debug to true if you want to inspect the dropdown
```
achekulaev: should it be set to false?

docs/layouts/partials/menu-footer.html
```diff
@@ -9,3 +9,12 @@
...
+ });
+</script>
```
achekulaev: newline missing

shelane: True. And the debug needs to be set back to false. Will do.

shelane: False is the default and I forgot that I had set it to true when trying to get the search results box to appear above the text of the documentation.

lmakarov:
There is something wrong with styles and/or index:

image

shelane:
@lmakarov Is there a good comparison site where I can see a working version that you know of?

lmakarov:
Not sure about good, but here are all of them:
https://github.com/algolia/docsearch-configs/tree/master/configs

Note: we can submit a PR to have our config updated if necessary:
https://github.com/algolia/docsearch-configs/blob/master/configs/docksal.json

shelane:
I took a look at [this site](https://djrad.herokuapp.com/) to see how the search is working. That seems what it should be. 
https://github.com/algolia/docsearch-configs/blob/master/configs/djrad.json

Is it something about the selectors in the docksal.json file? I was trying to find more documentation, but I only seem to find docs related to custom search versus this method.

shelane:
Here is the [relevant documentation](https://community.algolia.com/docsearch/config-file.html).

I know know if this has anything to do with the current problem or not, but still doesn't seem right:
```
"lvl0": {
      "selector": ".links > a:nth-child(2)",
      "global": true,
      "default_value": "Documentation"
},
```

And then lvl1 one is:
```
"lvl1": "#body-inner h1",
```

I think lvl1 needs to be moved to lvl0 and the rest moved up. Thoughts?

lmakarov:
@shelane Unfortunately, I cannot help much with the front-end stuff. We can try your suggestion and submit a PR. There is no staging with DocSearch, so I guess, all changes just go live once merged.

Note: if DocSearch integration is difficult to manage and becomes a burden, we could still go with a custom Algolia integration and manage the whole thing ourselves. It will be more work, but also give us more control and flexibility. Thoughts?

shelane:
Since we don't have a "production" version yet, I don't think it will hurt us to experiment. Let's try the PR first:
```
"selectors": {
    "lvl0": {
      "selector": "#body-inner h1",
      "global": true,
      "default_value": "Documentation"
    },
    "lvl1": "#body-inner h2",
    "lvl2": "#body-inner h3",
    "lvl3": "#body-inner h4",
    "lvl4": "#body-inner h5",
    "lvl5": "#body-inner h6",
    "text": "#body-inner p, #body-inner li"
},
```

shelane:
@lmakarov Found the conflict: Hugo learn theme defines a div with a class `searchbox` that surrounds the `#search-by` input. Algolia search uses the same class to define styles. Attempting to move the Algolia search css to the top and then resetting wasn't sufficient.

The solution:
To not conflict with the theme search, I had to create a partial layout for the search anyway to exclude the javascript of the theme search. So, I renamed the class to `searchboxed` and included the theme styles into the `theme-docksal.css` files. Now there is no conflict and it is behaving properly.

lmakarov:
@shelane thanks! Definitely looks much better now 👍 

Though, I noticed another problem with the search integration and the Learn theme. Apparently, you can use left/right keys to iterate through the documentation pages. This needs to be disabled when focus is on the search box input as this results in unexpected page reloads when you try to edit the text in the search box (moving the cursor left/right).

shelane:
Hmmm. That's interesting behavior. I'll take a look at that.

shelane:
Apparently this is not a conflict between Algolia and the Learn theme. This behavior exists in the Learn theme all by itself. If you check [the current docs site](https://docs.docksal.io), you will see the same problem. I am trying to see where this functionality is originating and how we might override that.

shelane:
I have added a jQuery function to prevent this unwanted functionality for input boxes and submitted an issue with the hugo-theme-learn repo. https://github.com/matcornic/hugo-theme-learn/issues/216

lmakarov:
Thanks! The new Algolia search widget looks and works great 🎉

PR #894, opened 2018-12-12
Support default Ed25519 SSH key

GMTA:
When using `ssh-keygen -t ed25519`, a file on the path `~/.ssh/id_ed25519` is created by default. This adds support to Docksal for that default key location.


achekulaev:
Thanks!

Issue #657, opened 2018-06-27
Create a check that directory is mapped in Docker


achekulaev:
For various reasons on Mac and Windows local directories may be not mapped to Docker, leading to issues like https://github.com/docksal/docksal/issues/338. 

The output is very confusing to users. They don't know what happens.

Would be valuable to check that directory is mapped inside Docker and provide recommendations for resolution dependent on the platform and vbox/native mode used.

Windows + VBox: possible SMB issues (described in troubleshooting)
Windows + D4W: possible [password change](https://github.com/docksal/docksal/issues/625#issuecomment-400752864).
Mac+VBox: possible NFS daemon or configuration issues
Mac+D4M: check that this folder is shared with Docker in settings

achekulaev:
The most straightforward and cross-platform way to do it, would be to run smallest possible container and mounting current dir

```bash
docker run -v `pwd`:`pwd` alpine sh -c "ls $(pwd)"
```

It is pretty expensive though. 1-1.5 seconds, when image is already cached.

achekulaev:
This ticket could allow returning `target:=run-cli` that was removed because we cannot guarantee that $HOME is mapped in Docker

achekulaev:
Related issue #850

achekulaev:
Ok it is not a perfect check since say `/tmp` will report as "mapped" only because it exists in Docker no matter what. Might be true for number of other folders but I don't expect it to be a big issue.

PR #902, opened 2018-12-18
Only mount custom certificates folder to the proxy if it is mapped to Docker (fixes #850)

### LINKED ISSUE #850: CONFIG_CERTS mount causes error during system reset
```
$ fin system reset
Restoring network settings...
Resetting Docksal services...
 ➔ proxy
docker: Error response from daemon: invalid mount config for type "bind": bind source path does not exist: /Users/leonid/.docksal/certs.
See 'docker run --help'.
 ERROR:  Failed starting the proxy service.
```

This also happens during update/install.

Introduced in Docksal [1.10.0](https://github.com/docksal/docksal/releases/tag/v1.10.0) via https://github.com/docksal/docksal/commit/c9bbadfbd24900e30bcc61a8203366efae9cd357
Specifically: https://github.com/docksal/docksal/blob/develop/bin/fin#L3364-L3366

The issue is the same as in #771 - we cannot rely on the user's home directory to be mounted/available  inside the VirtualBox VM.

achekulaev:
Check that certificates folder is mapped into Docker with new function introduced in #657. Fixes #850 

[ Squash merge when merging ]

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -3420,7 +3420,6 @@ install_proxy_service ()
 	local mount_certs
 	if [[ -d "${CONFIG_CERTS}" ]] && is_docker_path "${CONFIG_CERTS}"; then
 			mount_certs="--mount type=bind,src=${CONFIG_CERTS},dst=/etc/certs/custom"
-		fi
 	fi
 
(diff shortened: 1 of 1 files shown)


lmakarov:
Yeah... I guess I'm used to how this is handled in Bitbucket - via a checkbox when you merge the PR. Thanks for the reminder.

Issue #905, opened 2018-12-19
Rename web service declaration into apache


achekulaev:
`web` should be renamed to `apache` in `services.yml`. The old `web` declaration should remain in `services.yml` until November 2019.

shelane:
Use of the new service name should be changed in the documentation immediately with this new service name so that when it's removed, people won't be scrambling to make changes.

lmakarov:
Similar to https://github.com/docksal/docksal/pull/906#issuecomment-448432430, we will keep using `web` as the service name in the default stacks, but extend the services from the new `apache` service definition in `services.yml`.

I have to admit, that we ARE adding confusion with different service names in `services.yml` vs, e.g., `stack-default.yml`. We will probably need to revise docs and see how this can be explained in a clear way. Also, the fact that the images would still be names `docksal/web` (and `docksal/db`) may add more to the confusion...

lmakarov:
Yes, eventually we will have `nginx` added as a `web` service option.

lmakarov:
For `db` - we already have the `pgsql` option in `service.yml`.
Still need support for `fin db` with pgsql in `fin` (#567) and docs (#564).

PR #906, opened 2018-12-19
Remove deprecated `db` declaration

achekulaev:
Has no ticket.
Related ticket: #905
cc @sean-e-dietrich

### DIFF
--- a/stacks/services.yml
+++ b/stacks/services.yml
@@ -24,25 +24,6 @@ services:
       - ${DOCKSAL_DNS1}
       - ${DOCKSAL_DNS2}
 
-  # DB (legacy)
-  # TODO: Deprecated and to be removed in October 2018
-  db:
-    hostname: db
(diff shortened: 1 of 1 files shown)


shelane:
Documentation needs updating as well. I'm looking now... So would `fin project reset db` be changed to `fin project reset mysql` since it's the service name you are changing?

lmakarov:
I was also concerned with that, but with the way stacks are defined (see [stack-default.yml](https://github.com/docksal/docksal/blob/ac350de948983dceeb378c03ebe9035a2024da77/stacks/stack-default.yml#L15) and [stack-acquia.yml](https://github.com/docksal/docksal/blob/ac350de948983dceeb378c03ebe9035a2024da77/stacks/stack-acquia.yml#L22)) the resulting service name is still `db`, so nothing really changes from a user perspective.

PR #907, opened 2018-12-19
updates uses of db service in docs with mysql

shelane:
This updates documentation where the `db` service was used. I know that not every occurrence of `db` should be replaced with mysql, so please check that all the right places have been updated.


lmakarov:
Nothing changes from a user perspective, since we still use `db` as the service name in the actual stack definitions (default and acquia stacks). See https://github.com/docksal/docksal/pull/906#issuecomment-448432430

PR #914, opened 2018-12-21
Documenting image tags

### LINKED ISSUE #243: Docs: List available tag options for all services
We should list tags available for each service and explain what they are in the docs.

E.g. for [Apache Solr](http://docs.docksal.io/en/master/tools/apache-solr/) for only list `docksal/solr:1.0-solr4` as an option, however there is also `docksal/solr:1.0-solr3` available.
Also, need to mention stable releases vs edge ones.

shelane:
Closes #243

### DIFF
--- a/docs/content/core/overview.md
+++ b/docs/content/core/overview.md
@@ -43,15 +43,15 @@ Each project usually consists of at least 3 services: `web`, `db`, and `cli`.
 
 ### web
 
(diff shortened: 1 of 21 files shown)


achekulaev:
@lmakarov do you have an idea why build shows as failed?

shelane:
Looks like Netlify is having some issues: https://www.netlifystatus.com/

PR #916, opened 2018-12-23
Provider Integration

### LINKED ISSUE #912: Create 'pull' command
Should be able to do 3 things: pull db, files and code from Acquia and Pantheon at minimum (+Platform at best).

`fin pull db` - pull db using platform dependent tool(s)
`fin pull files` - pull files using platform dependent tool(s)
`fin pull code` - perform `git pull`
`fin pull` - perform all of the above

Platform to connect to configuration should be read from environment vars.

Push command will follow in the future release.

sean-e-dietrich:
Provides new command to interact with providers.

The providers refer to the following:

`acquia` - This refers to interacting with Acquia Cloud integration. Not all users will have access to this but can use it to download and interact with getting backups from Acquia. If the user does not have access to Acquia Cloud they will need to fall back to `drush` as the provider.

`pantheon` - This will interact with running terminus to pull files and db down.

`platformsh` - This will interact with the user's Platform.sh account.

`drush` - General Drupal Hosting with Drush support to pull files and/or database.

`wp` - General WordPress Hosting with SSH access to pull database currently.

Fixes #912


bin/fin
```diff
@@ -6261,6 +6343,630 @@ set_project_root_permissions ()
...
+		${rsync_options}
+	if_failed_error "Could not rsync files from Pantheon. Confirm that SSH Key has been added to Pantheon account and added to Docksal SSH-Agent."
+}
+
+# Pull process to pull db from Pantheon.
+# Pull from Pantheon backup system so that can pull the live db and not impact the site performance.
+provider_pull_pantheon_db ()
+{
+	# Get Time of Last Backup In Pantheon.
+	local last_backup=$(_exec -e COLUMNS="200" -e TERMINUS_HIDE_UPDATE_MESSAGE=1 -e TERMINUS_HIDE_GIT_MESSAGE=1 terminus backup:list ${HOSTING_SITE}.${HOSTING_ENV} --element=db --format=list --field=date | head -n 1)
```
achekulaev: What's is the reason of COLUMN=200 hardcode?

bin/fin
```diff
@@ -6261,6 +6343,630 @@ set_project_root_permissions ()
...
+	if [[ -d ${project_directory} ]]; then
+		echo-error "Project Directory ${project_directory} already exists."
+		exit 1
+	fi
+
+	provider_pull_find_function ${HOSTING_PLATFORM} 'init' 1 "Init function not currently supported for ${HOSTING_PLATFORM}."
+	echo-green "Pulling code complete."
+
+	# Change into the project directory.
+	cd ${project_directory}
```
achekulaev: needs quotes

bin/fin
```diff
@@ -6261,6 +6343,630 @@ set_project_root_permissions ()
...
+
+	# Query the provider for the given repo url. Only Pantheon, Platform.sh, and Acquia support this.
+	# All others will need to do a git clone.
+	if [[ "${ARGV[0]}" == "init" ]]; then
+		provider_pull_init
+		return
+	fi
+
+	# Check to see if providers require anything else.
+	provider_pull_find_function ${HOSTING_PLATFORM} 'precheck'
```
achekulaev: 1. Why is there 2 "check requirements" functions?
2. Precheck function in all provides seem to always use generic precheck
3. Why is precheck executed after check, which is semantically weird, since it's precheck should be the one executed first.

bin/fin
```diff
@@ -6261,6 +6343,630 @@ set_project_root_permissions ()
...
+	# Get the time from 24 hours ago.
+	local yesterday=$(_exec date --date='yesterday' "+%Y-%m-%d %H:%M:%S");
+
+	if [[ "${file_exists}" == "0" ]] || [[ -n "${file_time}" ]] || [[ "$FORCE" == "FORCE" ]]; then
+		echo-yellow "Cached DB file needs updating"
+
+		# Check to see if File Is Older than 1 hour
+		if [[ "$file_exists" > "0" ]] && [[ -z "${file_time}" ]] || [[ "$FORCE" == "FORCE" ]]; then
+			echo-yellow "Removing old database file"
+			_exec rm -rf ${db_file}
```
achekulaev: Needs quotes. Needs safety to avoid removing something beyond project root

bin/fin
```diff
@@ -6476,7 +6476,7 @@ provider_pull_db ()
 		# Check to see if File Is Older than 1 hour
 		if [[ "$file_exists" > "0" ]] && [[ -z "${file_time}" ]] || [[ "$FORCE" == "FORCE" ]]; then
 			echo-yellow "Removing old database file"
-			_exec rm -rf ${db_file}
+			_exec rm -rf "${db_file}"
```
achekulaev: Please also take a look at https://github.com/docksal/docksal/blob/develop/bin/fin#L6116-L6118

In that spot I have a simple, but at least some protection from someone tricking users into setting malicious env vars that would make `fin` perform `rm -rf` on a wrong place!

Also since you are only deleting a folder please use `rm -f`.

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+	if [[ "${HOSTING_PLATFORM}" == "" || "${HOSTING_SITE}" == ""  ]]; then
+		echo-error "Hosting Provider not set up. Following variables must be set: HOSTING_PLATFORM and HOSTING_SITE"
+		exit 1
+	fi
+
+	HOSTING_PLATFORM=$(echo ${HOSTING_PLATFORM} | tr '[:upper:]' '[:lower:]')
+
+	# Check to see if a support platform is accepted.
+	if [[ "${HOSTING_PLATFORM}" != "acquia" ]] && [[ "${HOSTING_PLATFORM}" != "pantheon" ]] && [[ "${HOSTING_PLATFORM}" != "platformsh" ]] &&
+		[[ "${HOSTING_PLATFORM}" != "drush" ]] && [[ "${HOSTING_PLATFORM}" != "wp" ]]; then
```
achekulaev: Can be simplified. See example https://github.com/docksal/docksal/blob/1d99f684d7dc4006453c09f74047703eb4dc4b6b/bin/fin#L6508

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
 	is_root && chown -R 1000:1000 "$PROJECT_ROOT"
 }
 
+# Pull main command for interacting with Hosting Providers.
```
achekulaev: Please describe params

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+	# Execute a pull on Files.
+	if [[ "${ARGV[0]}" == "files" ]] || [[ "${ARGV[0]}" == "" ]]; then
+		provider_pull_files
+	fi
+
+	echo-green "Pull Completed"
+}
+
+# Provider Pull Init Subcommand
+# Used specifically for pulling the repo from a provider.
```
achekulaev: Please describe params

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+	fi
+}
+
+# Provider Pull Code Subcommand
+# Used specifically for pulling the repo from a provider.
+provider_pull_code ()
+{
+	echo-green "Pulling code"
+	provider_pull_find_function ${HOSTING_PLATFORM} 'code' 1
+	echo-green "Code Pull Successful"
```
achekulaev: Error checking required or at least `#TODO: implement error checking`

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+	fi
+
+	# ToDo: Do we need to account for Multi-Site and other DB Names?
+
+	# Import the Database.
+	# Truncate the existing db.
+	echo-yellow "Truncating Existing Database"
+	mysql_db_truncate "${DB_NAME}" > /dev/null
+	# Execute ZCAT on the db file and import into mysql database.
+	_exec "zcat < ${db_file} | mysql -u ${__db_user} -p${__db_password} -h db ${__db_database}"
```
achekulaev: should `mysql_import` import function be used?

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+	local rsync_options="${RSYNC_OPTIONS} --exclude='styles' --exclude='css' --exclude='js'"
+	provider_pull_find_function ${HOSTING_PLATFORM} 'files' 1
+	echo-green "File Pull Successful"
+}
+
+##############################
+### GENERIC PROVIDER FUNCTIONS
+##############################
+
+# Generic function for searching if function exists and executing.
```
achekulaev: Please describe params

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+	if [ $? -eq 0 ]; then
+		# Execute function if found
+		${pull_function}
+	else
+		# If this was required and not found execute error message.
+		if [[ -n $required ]]; then
+			echo-error ${required_message}
+			exit 1
+		fi
+	fi
```
achekulaev: This if-else can be simplified

```bash
type "${pull_function}" >/dev/null 2>&1
# If this was required and not found execute error message.
if_failed_error "${required_message}"
# Execute function if found
${pull_function}
```

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+provider_pull_generic_drush_rsync ()
+{
+	echo-yellow "Downloading files from ${HOSTING_SITE} on the ${HOSTING_ENV} environment"
+	[[ ! -d ${PROJECT_ROOT}/${DOCROOT}/${FILES_DIR} ]] && mkdir -p ${PROJECT_ROOT}/${DOCROOT}/${FILES_DIR} > /dev/null
+	_exec "drush -y rsync \
+		@${HOSTING_SITE}.${HOSTING_ENV}:%files \
+		${PROJECT_ROOT}/${DOCROOT}/${FILES_DIR} \
+		--delete \
+		${rsync_options}"
+	if_failed_error "Error Downloading Files from ${HOSTING_SITE} on the ${HOSTING_ENV} environment"
```
achekulaev: Downloading files might not be that big of an issue and maybe it makes sense to just `echo-error` here rather that fail and exit script.

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+	# Following will grab all databases for a site.
+	# local acquia_dbs=$(_exec drush "@${HOSTING_SITE}.${HOSTING_ENV}" ac-database-list | awk '{print $3}')
+
+	local acquia_db_name="${ARGV[1]}"
+	if [[ -z "${acquia_db_name}" ]]; then
+		echo-error "Database name is required."
+		exit 1
+	fi
+
+	local query=$(_exec drush @${HOSTING_SITE}.${HOSTING_ENV} ac-database-instance-backup-list ${acquia_db_name});
```
achekulaev: it should be split into 2 lines. Bash has a quirk that this line will always return result `0` even if command fails! That is because the last command will be `local`.

You can reproduce it yourself

```bash
#!/bin/bash

testie () {
    local abc=$(exit 1)
    echo $?
}

testie
```
So the line needs to be changed to:

```bash
local query
query=$(...)
```

You only need to do the separation if you need the correct error code.

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+	# Set the default db_name just incase we need to cache this for Acquia or other providers.
+	local db_name=${ARGV[1]:-"default"}
+	# Set the db file that will be stored within the cli container
+	local db_file="/tmp/${db_name}.${HOSTING_ENV}.${HOSTING_SITE}.sql.gz"
+	# Check to see if the file exists in the cli container
+	local file_exists=$(_exec "ls -1 /tmp/**  2> /dev/null | grep \"${db_file}\"\$ | wc -l | tr -d '[:space:]'")
+	# Get the time of that file within the cli container.
+	local file_time=$(_exec "find ${db_file} -mmin +360 2> /dev/null | tr -d '[:space:]'")
+	# Get the time from 24 hours ago.
+	local yesterday=$(_exec date --date='yesterday' "+%Y-%m-%d %H:%M:%S");
```
achekulaev: Wouldn't it be more robust to compare dates in unix timestamp format? `date --date='yesterday' +%s`

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+}
+
+# Init process to pull code from Pantheon
+provider_pull_pantheon_init ()
+{
+	# If HOSTING_ENV isn't set default to dev.
+	HOSTING_ENV=${HOSTING_ENV:-dev}
+	# Set the BRANCH to be what the HOSTING_ENV is.
+	local repo_checkout=${HOSTING_ENV}
+	local repo_url=$(run_cli -e COLUMNS="200" "terminus connection:info ${HOSTING_SITE}.dev --field=\"Git Command\" | awk '{print \$3}'")
```
achekulaev: As before break `local var` and assignment to it into 2 lines

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+	run_cli "git clone -q --depth=1 --branch='${repo_checkout}' ${repo_url} ${project_directory}"
+	if_failed_error "Error cloning Pantheon project ${HOSTING_SITE} - ${HOSTING_ENV} in ${project_directory}"
+}
+
+# Pull process to pull files from Pantheon.
+# Need to run a custom rsync command as every environment is different.
+provider_pull_pantheon_files ()
+{
+	# Get the SFTP Hostname from Site's Info.
+	local host_url=$(_exec -e COLUMNS="200" "terminus connection:info ${HOSTING_SITE}.${HOSTING_ENV} --field=\"SFTP Command\" 2>/dev/null | awk '{print \$4}'")
```
achekulaev: Break into 2 lines

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+		${rsync_options}
+	if_failed_error "Could not rsync files from Pantheon. Confirm that SSH Key has been added to Pantheon account and added to Docksal SSH-Agent."
+}
+
+# Pull process to pull db from Pantheon.
+# Pull from Pantheon backup system so that can pull the live db and not impact the site performance.
+provider_pull_pantheon_db ()
+{
+	# Get Time of Last Backup In Pantheon.
+	local last_backup=$(_exec -e COLUMNS="200" terminus backup:list ${HOSTING_SITE}.${HOSTING_ENV} --element=db --format=list --field=date | head -n 1)
```
achekulaev: Break into 2 lines

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+	HOSTING_ENV=${HOSTING_ENV:-master}
+	echo-yellow "Cloning Platform.sh Site Locally: ${HOSTING_SITE} - ${HOSTING_ENV} in ${project_directory}"
+	run_cli "platform project:get -q --environment=${HOSTING_ENV} ${HOSTING_SITE} ${project_directory}"
+	if_failed_error "Could not communicate with Platform.sh. Check connection and SECRET_PLATFORMSH_CLI_TOKEN is set to an active token."
+}
+
+# Pull process to pull files from Platform.sh
+provider_pull_platformsh_files ()
+{
+	local host_url=$(_exec "platform ssh --pipe --project=${HOSTING_SITE} --environment=${HOSTING_ENV} 2>/dev/null");
```
achekulaev: Break into 2 lines

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+# Pull process to pull files from WP-CLI.
+provider_pull_wp_code ()
+{
+	# Execute a generic git pull on code.
+	provider_pull_generic_git_pull
+}
+
+# Execute specific functions if they exist within the hooks directory.
+# Examples could be hook_init_pre or hook_db_import.
+# Specifically look for hooks to run pre and post actions.
```
achekulaev: Please describe params

bin/fin
```diff
@@ -6287,6 +6344,631 @@ set_project_root_permissions ()
...
+	if [[ "${file_exists}" == "0" ]] || [[ -n "${file_time}" ]] || [[ "$FORCE" == "FORCE" ]]; then
+		echo-yellow "Cached DB file needs updating"
+
+		# Check to see if File Is Older than 1 hour
+		if [[ "$file_exists" > "0" ]] && [[ -z "${file_time}" ]] || [[ "$FORCE" == "FORCE" ]]; then
+			echo-yellow "Removing old database file"
+			_exec rm -rf "${db_file}"
+		fi
+
+		echo-yellow "Pulling new database file..."
```
achekulaev: Here and in other places where you do `echo-yellow`, but it is not a warning. I understand you want to separate sub-streps, but yellow is really mostly a warning (unless single words are highlighted). 

So using color like `acqua` might work better.
Or maybe add and use bold "color"

```
bold="\033[1m"
```

bin/fin
```diff
@@ -4747,7 +4803,7 @@ run_cli ()
 	done
 
 	# Set default image
-	local RUN_IMAGE="${image:-docksal/cli:2.5-php7.1}"
+	local RUN_IMAGE="${image:-docksal/cli:2.6-php7.2}"
```
achekulaev: @lmakarov so I propose we replace this with the new `cli:2-php7.2` tag. Cases like this were the main reason for its creation. Agreed or any comments?

sean-e-dietrich: I'm fine with this and agree with it. We are running into issues where a different version is constantly being used all the time.

bin/fin
sean-e-dietrich: I wish I could only the issue is that need the command to run in cli as that is where the db file is stored.

bin/fin
sean-e-dietrich: Found a way around this.

sean-e-dietrich:
@lmakarov @achekulaev @shelane i think we are ready to review this. Docs have been updated. Tests are passing.

Issue #927, opened 2019-01-04
Update Coding Standard Docs


sean-e-dietrich:
`docksal/service-cli` now includes WP Coding Standards after https://github.com/docksal/service-cli/pull/88

We should include documentation on https://docs.docksal.io/tools/phpcs/ to include information that this exists. 

Additionally the docs should also reflect that `Drupal` standard should now be changed to `Drupal,DrupalPractice`

sean-e-dietrich:
Closed via https://github.com/docksal/docksal/commit/6a659bf1199a4597806806316613d62b265f7d23

PR #929, opened 2019-01-04
Updated documentation for PHPCS to include WordPress

sean-e-dietrich:
Added information to the docs for WordPress PHPCS additionally made commands for easier downloading.

Solves https://github.com/docksal/docksal/issues/927

### DIFF
--- a/docs/content/tools/phpcs.md
+++ b/docs/content/tools/phpcs.md
@@ -5,7 +5,8 @@ aliases:
 ---
 
 
-`phpcs` and `phpcbf` are readily available in the cli container.
+`phpcs` and `phpcbf` are readily available in the cli container. Both [Drupal Coding Standards](https://www.drupal.org/docs/develop/standards) and [WordPress Coding Standards](https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards) 
+have been installed and are ready to use.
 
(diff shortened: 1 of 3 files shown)


shelane:
By creating separate example commands for Drupal and WP, should the existing `examples/.docksal/commands/phpcs` be removed?

sean-e-dietrich:
@shelane it actually should have been removed

@achekulaev @lmakarov are you ok with it being removed and replaced with different examples as long as it's documented in the docs properly?

achekulaev:
I'm ok with renaming. However for visual consistency I would rather call binaries `phpcs_*` instead of `*_phpcs`.

lmakarov:
I believe both `phpcs_drupal` and `phpcs_wordpress` belong to the [addons](https://github.com/docksal/addons) repo. `phpcs` should probably be also removed from examples in this repo.

achekulaev:
@lmakarov yes and no. Since the ancient Docksal times it served as an example of the folder structure.  If we remove the we need to be removing examples folder altogether (it contains ancient images in docksal.env there by the way)

achekulaev:
@shelane this PR is assigned to you, others reviews are optional, so whenever you feel it's ready you can merge.

PR #955, opened 2019-02-08
Changing VSCode docs to reflect working launch.json.

JDDoesDev:
The recently committed documentation for VSCode Xdebug did not work on my install.  I have successfully been using Xdebug with VSCode for some time with these settings, so I recommend that we use this example instead of the current one.


achekulaev:
@lmakarov do you have an idea why tests are failing on setting ENV variables?

sean-e-dietrich:
@achekulaev I believe secrets are failing because it's a forked repo with the PR and the variables are marked as secret.

lmakarov:
Tests will always fail with PRs since there are private ENV variables set in Travis, which are used in tests and which are not passed to PR builds. We'll need to refactor the builds/tests to only run PR-compatible tests on PR builds.

sean-e-dietrich:
@JDDoesDev you are correct. Thanks!

PR #960, opened 2019-02-15
vm functions refactoring

lmakarov:
- Dropped unused vm use cases
  - `vm create`
  - `vm active`
  - `vm ip <IP>`
- Deprecated `fin vm start/stop` in favor of `fin system start/stop`

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -774,7 +774,7 @@ check_docker_running ()
 
 	if [[ ${docker_status} -eq 255 ]] && (! is_docker_native && ! is_wsl); then
 		echo-yellow "It looks like '$DOCKER_MACHINE_NAME' docker machine is not running."
-		_confirm "Run 'fin vm start' to start it now?"
+		_confirm "Start it now?"
 		docker_machine_start
 		if_failed "Could not start Docksal docker machine properly"
 		# re-check status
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -2143,28 +2142,22 @@ docker_machine_create ()
 	if [[ $? -eq 0 ]]; then
 		DOCKER_MACHINE_NAME="$machine_name"
 		DOCKER_MACHINE_STATUS='Running'
-		vm active "$machine_name"
```
achekulaev: Don't remove this. If user had another vm active before Docksal this is needed.

bin/fin
```diff
@@ -3156,35 +3065,41 @@ system_status ()
...
 
 # Stops/disables all Docksal things
 system_stop ()
 {
 	if ! is_linux && ! is_docker_native && ! is_wsl; then
-		echo-green "Stopping Docksal VM..."
-		# VM stop will perform configure_resolver off
-		vm stop
+		# Stop the docker-machine VM.
+		if is_docker_machine_running; then
```
achekulaev: KISS principle says there is no need for this check here. Because if it is stopped docker-machine will handle it itself, but at the same time is there is some bug with machine status detection it can prevent this from stopping the machine. 
So no gains, and a loss in possible bug prevention. 
I would leave this check out.

lmakarov:
Replaced by #989

PR #962, opened 2019-02-17
[WIP] SSH Proxy Feature

sean-e-dietrich:
Added feature that would allow integration with the [SSH Proxy](https://github.com/docksal/service-ssh-proxy). This comes in hand with Sandboxes and allowing for SSHing into a specific project without the need to specifically ssh into the host machine. This also comes in handy when wanting to use drush remotely.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -261,6 +261,7 @@ URL_WINPTY="https://github.com/rprichard/winpty/releases/download/${REQUIREMENTS
 
 IMAGE_SSH_AGENT=${IMAGE_SSH_AGENT:-docksal/ssh-agent:1.2}
 IMAGE_VHOST_PROXY=${IMAGE_VHOST_PROXY:-docksal/vhost-proxy:1.4}
+IMAGE_SSH_PROXY=${IMAGE_SSH_PROXY:-docksal/ssh-proxy:edge}
 IMAGE_DNS=${IMAGE_DNS:-docksal/dns:1.1}
 
 #---------------------------- Helper functions --------------------------------
@@ -994,6 +995,7 @@ _start_containers ()
 	docker-compose up -d --remove-orphans --build || return 1
 	_healthcheck_wait
 	_vhost_proxy_connect
+	_ssh_proxy_connect
 }
 
 # @param $1 "-a" || "--all"
@@ -1041,6 +1043,12 @@ _stop_containers ()
 		return
 	fi
 
+	if [[ "$1" == "ssh-proxy" ]] ; then
+		echo-green 'Stopping Docksal ssh-proxy service...'
+		docker stop docksal-ssh-proxy >/dev/null
+		return
+	fi
+
 	load_configuration
 	check_project_unique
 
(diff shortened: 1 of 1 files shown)


christianwiedemann:
Hi @sean-e-dietrich,

 I created an SSH server with go to access Docksal container. Maybe this can be an alternative (or an starting point of docksal/service-ssh). https://github.com/andock/ssh2docksal

I think it has the two advantages. First it works without an ssh server installed on the container. The second one is that the authentification could be easily extend to support differnt backends. 
The main disadvantage from my point of view is that the sftp implementation is much slower than direct access. 

Maybe it makes sense to install an ssh service also for development and not only for sanboxes. 
This could help to remove port forwarding and the ssh server on the container. 

I am using forwarded ports right now to access the database and to phpcs / remote debugging with phpStorm. I think it would be much more straight forward to access the container with username.

christianwiedemann:
hm, and shouldn't be such a service part of docksal? Maybe it makes more sense to add this functionality as an global addon. Somthing like 
```
fin addon install -g ssh-server
```
Or maybe an unified mechanismus to install global container. This could help to replace/config an global service.

sean-e-dietrich:
Hi @christianwiedemann I'm not sure I am following. What you have does look pretty amazing though, and think that it would be a possible addon.

I want to point out that we started with the following https://github.com/docksal/service-ssh-proxy

There are plans in the future to do more with it but right now it gives the ability to manage keys on a per project basis for added security in the sandbox mode, but also gives you the ability to have global keys that can be used on all projects. There are also additional features where it will look for only docksal projects and there are plans in the future to be able to only include projects or exclude projects you want.

I'm also happy to hear it's written in GO because I personally cannot write GO but was looking for a library that would work and a similar project https://github.com/tg123/sshpiper was also written in GO. This is what the service-ssh-proxy project is based off of.

christianwiedemann:
@sean-e-dietrich Sorry for the confusion. I think I mixed two topics. 
The first one was to use my ssh implementation but I understand sshpiper is the realy the better solution. 

The second point is that I think it would be realy cool to have the posibility to enable/disable/config services. Right now all services are hard coded inside Docksal. This makes it unpossible to disable for example the dns service in sandbox mode for example or replace the http proxy with another Service or add an complete new service like ssh server or maybe a logging service. And if I don't need the service there is no need to activate it. So I think an SSH service is also very handy for development but I am sure the most people don't need it. 

To sum up: It would be cool to have something like the addons for services and Docksal installs it default services also with this mechanism. Maybe there is global docker-compose.yml which is started and this docker-compose file is managed through fin.

sean-e-dietrich:
@christianwiedemann understand now.

Just to clarify you can control these things. There are different variables that can be set within your `$HOME/.docksal/docksal.env` file that will give you the ability to change and to set these.

Example:
These are the main images you are referring to. Now,

`IMAGE_VHOST_PROXY`
`IMAGE_SSH_AGENT`
`IMAGE_DNS`

Now, the problem is we cannot guarantee that some of the automation will continue to work. Example would be commands related to those images. Adding the SSH Keys to the Agent, or anything else. Granted that would be if you entirely chose to swap out some of the items there.

Heads up also these variables and others should be documented here https://docs.docksal.io/stack/configuration-variables/

PR #969, opened 2019-02-22
adds notice for moving a project on host

### LINKED ISSUE #948: Possible to move project directory to another location in the filesystem?
For example, say I have a Docksal project at `~/sites/abc` and I decide to move it to `~/stuff/abc`, then when I `fin start` in the new location, it complains that project ABC already exists.

Is there a way to move projects from one place to another while keeping the same directory name, without having to `fin rm` first?

shelane:
Closes #948


mikecrittenden:
Thanks @shelane, this is perfect.

PR #975, opened 2019-02-26
fixes css for better positioning

shelane:
CSS fixes for pages where the content does not go full length.


lmakarov:
@shelane can you please provide before and after screenshots or steps to verify?

shelane:
Before:
before

After:
after

Issue #974, opened 2019-02-25
Docksal aliases don't work within commands


charginghawk:
**Description**

If you try to use a docksal alias within a docksal command, it fails.

This makes it hard to automate cross site interactions (at the most basic level, initializing one project as a dependency of another).

**Steps to reproduce the issue:**

1. Create a test command like .docksal/commands/alias-within-command:

```
#!/usr/bin/env bash
fin @target-project config
```

2. Run command and get these results:

```
---------------------
COMPOSE_PROJECT_NAME_SAFE: current-project
COMPOSE_FILE:
/Users/user/.docksal/stacks/volumes-nfs.yml
/Users/user/Projects/current-project/.docksal/docksal.yml
/Users/user/Projects/current-project/.docksal/docksal-local.yml
ENV_FILE:
/Users/user/Projects/target-project/.docksal/docksal.env

PROJECT_ROOT: /Users/user/Projects/target-project
DOCROOT: docroot
VIRTUAL_HOST: target-project.docksal
VIRTUAL_HOST_ALIASES: *.target-project.docksal
IP: 192.168.64.100
MYSQL:

Docker Compose configuration
---------------------
ERROR: build path /Users/user/Projects/target-project/.docksal/web either does not exist, is not accessible, or is not a valid URL.
```

**Describe the results you received:**

The project association is pretty scrambled.

**Describe the results you expected:**

The command successfully runs for the targeted environment (as happens running the command outside of another docksal command).

achekulaev:
@charginghawk show output of `fin alias list`

achekulaev:
What makes you think that alias does not work? As far as I see everything works. 

You have custom Dockerfile in one of your projects that has relative build path, that is not resolved properly because you use aliases.

```
ERROR: build path /Users/user/Projects/target-project/.docksal/web either does not exist, is not accessible, or is not a valid URL.
```

You need to be mindful about relative paths when using aliases.

charginghawk:
@achekulaev I've gone through and made a more reproducible scenario, and this is an issue. I can't start another project from within a command, and the config output is different as well.

I've created two directories, `~/Projects/test1` and `~/Projects/test2`. I've cd'd into both and run `fin init` (so barebones, no changes, no code, default setups). With those two projects initialized, I've created two commands in test1:

```
$ cat .docksal/commands/test2-config 
#!/usr/bin/env bash
fin @test2 config

$ cat .docksal/commands/test2-start 
#!/usr/bin/env bash
[...]

PR #976, opened 2019-02-26
Added ability to run an alias in a custom command

sean-e-dietrich:
When running commands in custom commands with aliases there is an issue where the VIRTUAL_HOST and the PROJECT_SAFE_NAME

To reproduce:

* create two projects (project1 and project2)
* start both projects
* in project2 create a custom command in project2
```bash
#!/usr/bin/env bash

fin @project1 project status
```
* in project2 run `fin test`
* Should get a similar response

```
 ERROR:  Another project is already using the name 'project2'
         Change the name of the current project by renaming the project folder (folder name defines the project name and has to be unique)
         or remove the other project's stack by running fin project remove in /Users/sean.dietrich/Desktop/docksal_test/project2
```

Solves https://github.com/docksal/docksal/issues/974

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -6323,6 +6323,7 @@ addon_execute ()
 		# Run the command
 		CONTAINER_NAME="$_exec_target" _exec "$cmd"
 	else
+		export ADDON_SCRIPT=true
 		exec "$addon" "$@"
 	fi
 }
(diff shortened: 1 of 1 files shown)


achekulaev:
Thanks and I liked the name of the branch :)

PR #987, opened 2019-02-28
Adjusted Pull Command Variables

sean-e-dietrich:
Converted all provider pull variables to be lowercase and conform with other docksal options.

Also update documentation to account for these changes.


sean-e-dietrich:
@lmakarov thanks! So added variable in travis `RUN_PULL_TESTS` that when set will run the pull tests.

achekulaev:
@lmakarov Huston, we have a problem

Travis:
> The job exceeded the maximum time limit for jobs, and has been terminated.

lmakarov:
> Houston, we have a problem

Well, as I already mentioned, I believe `pull` commands should be moved into the `docksal/addons` repo and as separate addons, that can be tested independently. Might even need separate repos altogether for each of the hosting integrations.

achekulaev:
Any other way to prolong execution time?

sean-e-dietrich:
I'm thinking of cutting out the wp commands for now. Also all of starting/stopping for projects adds on probably about a couple of min for 29 tests. I added a variable that we test run_pull_tests I'm wondering if that is set then we don't run other tests. Similar to how the quicktests are set.

achekulaev:
https://docs.travis-ci.com/user/speeding-up-the-build

achekulaev:
@sean-e-dietrich nice work!

PR #990, opened 2019-03-02
Fixing Numbering on chapter documentatoin.

sean-e-dietrich:


sean-e-dietrich:
good catch @shelane ! Didn't see the dev notes. Updated index for use cases.

PR #993, opened 2019-03-06
`exec` refactoring

### LINKED ISSUE #942: Docksal Support for Alpine Images
Right now fin seems to have a hard dependency on `bash` for many operations in the containers to function.  This means that when running containers (like Alpine based images) you will run into many different issues with various fin commands. `fin bash` and `fin exec` being the 2 most obvious commands.  There are work arounds for some of these things by falling back to docker-compose, but that limits the benefits of fin greatly. `fin dc run <container> /bin/ash` for example will get you to a shell in an Alpine based image.

`node:10.15.0-alpine` was the specific image I was testing with from https://hub.docker.com/_/node/ for reference.

achekulaev:
`exec` and `bash` will use sh by default and only use bash for `cli` and `db` (complex db functions input requires `bash` as well).

Closes #942 
Relates to #924

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1463,11 +1463,13 @@ show_help_list_addons_in_help ()
 show_help_exec ()
 {
 	echo
-	echo "Execute command or file in \`cli\` service container."
+	echo "Execute commands or script in \`cli\` service container,"
+	echo "or execute commands in other containers when specified in params."
 	echo "fin exec will automatically cd into the same folder inside \`cli\`."
 
 	echo
 	echo "Usage: exec [-T] [--in=name] <command | file>"
+	echo "  [!] Parameters order matters."
 
 	echo
 	echo "Options:"
--- a/bin/fin
+++ b/bin/fin
@@ -4732,7 +4732,7 @@ _exec ()
 	local SHELL_NONINTERACTIVE="sh -lc"
 
 	# For cli use bash
-	if [[ "$CONTAINER_NAME" == "cli" ]]; then
+	if [[ "$container_name" == "cli" ]] || [[ "$container_name" == "db" ]]; then
 		SHELL_INTERACTIVE="bash -ilc"
 		SHELL_NONINTERACTIVE="bash -lc"
 	fi
--- a/bin/fin
+++ b/bin/fin
@@ -4733,7 +4733,7 @@ _exec ()
 	local SHELL_INTERACTIVE="sh -ilc"
 	local SHELL_NONINTERACTIVE="sh -lc"
 
-	# For cli use bash
+	# For cli and db use bash (complex SQL queries fail when fed from stdin to sh)
 	if [[ "$container_name" == "cli" ]] || [[ "$container_name" == "db" ]]; then
 		SHELL_INTERACTIVE="bash -ilc"
 		SHELL_NONINTERACTIVE="bash -lc"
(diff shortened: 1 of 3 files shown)


bin/fin
```diff
@@ -4711,66 +4719,75 @@ _exec ()
...
+	else
+		container_name="cli"
 	fi
 
-	CONTAINER_NAME=${CONTAINER_NAME:-cli}
-	container_id=$(get_container_id "$CONTAINER_NAME")
+	container_id=$(get_container_id "$container_name")
+
+	# Use sh as default shell
+	local SHELL_INTERACTIVE="sh -ilc"
```
lmakarov: Use lowercase variable names for local variables, to keep things consistent

bin/fin
```diff
@@ -4711,66 +4719,75 @@ _exec ()
...
+	local SHELL_NONINTERACTIVE="sh -lc"
+
+	# For cli use bash
+	if [[ "$container_name" == "cli" ]] || [[ "$container_name" == "db" ]]; then
+		SHELL_INTERACTIVE="bash -ilc"
```
lmakarov: Use lowercase variable names for local variables, to keep things consistent

bin/fin
```diff
@@ -4711,66 +4719,75 @@ _exec ()
...
+	# Add space if cmd is not empty
+	[[ $cmd != "" ]] && cmd="$cmd "
 	# Escape spaces that are "spaces" and not parameter delimiters (i.e., param1 param2\ with\ spaces param3)
 	if [[ $2 != "" ]]; then
-		cmd="$cmd "$(printf " %q" "$@")
+		cmd="$cmd"$(printf " %q" "$@")
 	# Do not escape spaces if there is only one parameter (e.g., fin run "ls -la | grep txt")
 	else
-		cmd="$cmd $@"
+		cmd="$cmd""$@"
```
lmakarov: `cmd="${cmd}${@}"` would be a cleaner option

docs/content/fin/fin-help.md
```diff
@@ -1,5 +1,5 @@
 ---
-title: "All commands reference"
+title: "fin help"
```
lmakarov: Keep the title "All commands reference"

bin/fin
```diff
@@ -4732,7 +4732,7 @@ _exec ()
 	local SHELL_NONINTERACTIVE="sh -lc"
 
 	# For cli use bash
```
lmakarov: Add a comment around why bash has to be used for `db`

bin/fin
```diff
@@ -4679,20 +4681,26 @@ image_registry_list ()
...
 	if ! is_tty; then
-		echo "not a tty"
+		echo-error "tty is required to start an interactive shell"
 		return 1
 	fi
 
-	# Pass container name to _run
-	CONTAINER_NAME=$1 _exec bash -il
+	# Use sh by default
+	if [[ "$1" != "" ]]; then
```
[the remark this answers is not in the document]
achekulaev: It is bash function it does not accept --in parameter

bin/fin
achekulaev: It is a constant. Hence uppercase

docs/content/fin/fin-help.md
achekulaev: It is the title that you set in help generation script. I manually changed it in another commit but the script replaced it again.

bin/fin
achekulaev: done

bin/fin
achekulaev: done

docs/content/fin/fin-help.md
lmakarov: Right. Let's then leave it as set in `scripts/fin-help-generate.sh` - "fin help"

PR #1016, opened 2019-03-28
Feature: Enable creation of default commands for command groups

dnotes:
As a project manager / lead developer, I would like to give my dev team a group of commands related to behat, but still allow them to type `fin behat` as a basic command. In order to do this, I need to be able to create a default command that will be executed when the name of the folder is typed. For example, I might have the following commands:

| Command | Runs something like ... |
| --- | --- |
| `fin behat` | `vendor/bin/behat "$@"` |
| `fin behat/info` | `vendor/bin/behat -d i --no-colors > available-tests.txt` |

To do this, I would make a folder structure as follows:

```
.docksal
  commands
    behat
      .default
      info
```

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -528,7 +528,7 @@ get_command_script ()
 	# Then search global docksal commands directory for the command
 	[ ! -f "$command_script" ] && command_script="$HOME/$DOCKSAL_COMMANDS_PATH/$1"
 	# Then search for a default command in a global command group
-	[ ! -f "$command_script" ] && command_script="$HOME/$DOCKSAL_COMMANDS_PATH/$1/.default"
+	[ ! -f "$command_script" ] && command_script="$HOME/$DOCKSAL_COMMANDS_PATH/$1/$1"
 	# If command was found then return it
 	[ -f "$command_script" ] && echo "$command_script"
 }
--- a/bin/fin
+++ b/bin/fin
@@ -527,6 +527,8 @@ get_command_script ()
 	[ ! -f "$command_script" ] && command_script="$COMMANDS_ROOT/$1/.default"
 	# Then search global docksal commands directory for the command
 	[ ! -f "$command_script" ] && command_script="$HOME/$DOCKSAL_COMMANDS_PATH/$1"
+	# Then search for a default command in a global command group
+	[ ! -f "$command_script" ] && command_script="$HOME/$DOCKSAL_COMMANDS_PATH/$1/.default"
 	# If command was found then return it
 	[ -f "$command_script" ] && echo "$command_script"
 }
--- a/bin/fin
+++ b/bin/fin
@@ -524,7 +524,7 @@ get_command_script ()
 	# First search project commands folder for the command
 	command_script="$COMMANDS_ROOT/$1"
 	# Then search for a default command in a group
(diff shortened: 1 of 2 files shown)


achekulaev:
Hi

Looks good to me except the naming convention. 

Currently addons made in a way that the "default" command is the file named the same as the folder (e.g. `.docksal/addons/mailhog/mailhog` will get run when `fin mailhog` command is run).

While this PR calls for different naming convention with `.default`.

To keep things the same let's keep the naming convention the same for commands, i.e. instead of `.docksal/commands/behat/.default` let's have `.docksal/commands/behat/behat` to run by default.

bin/fin
```diff
@@ -523,8 +523,12 @@ get_command_script ()
 	COMMANDS_ROOT="$(get_project_path)/$DOCKSAL_COMMANDS_PATH"
 	# First search project commands folder for the command
 	command_script="$COMMANDS_ROOT/$1"
+	# Then search for a default command in a group
+	[ ! -f "$command_script" ] && command_script="$COMMANDS_ROOT/$1/.default"
```
```suggestion
	[ ! -f "$command_script" ] && command_script="$COMMANDS_ROOT/$1/$1"
```

bin/fin
```diff
@@ -523,8 +523,12 @@ get_command_script ()
...
 	# Then search global docksal commands directory for the command
 	[ ! -f "$command_script" ] && command_script="$HOME/$DOCKSAL_COMMANDS_PATH/$1"
+	# Then search for a default command in a global command group
+	[ ! -f "$command_script" ] && command_script="$HOME/$DOCKSAL_COMMANDS_PATH/$1/.default"
```
```suggestion
	[ ! -f "$command_script" ] && command_script="$HOME/$DOCKSAL_COMMANDS_PATH/$1/$1"
```

docs/content/fin/custom-commands.md
```diff
@@ -176,7 +176,14 @@ services:
 
 ## Grouping Commands
 
-Docksal allows for commands to be grouped together within folders. This is particulary useful when creating a toolkit to share with other developers. Commands can be grouped within the Global Scope `~/.docksal/commands` and on a per project basis.
+Docksal allows for commands to be grouped together within folders. This is particulary useful when creating a toolkit to share with other developers. Commands can be grouped within the Global Scope `~/.docksal/commands` and on a per project basis. Command groups can also have a `.default` command that runs when the group name is given. For example, a project's `.docksal/commands` folder might contain a `test` directory as a command group:
```
achekulaev: Update documentation to describe naming convention proposed above

dnotes:
The only issue I see with this naming convention is when you list the commands (e.g. with `fin`) and they show up as `folder/command`; if the command name is .default then it's pretty obvious that you can just type the folder name, but with e.g. `behat/behat` it looks like you have to use the whole thing. So we should replace something from $cmd_name, but that kinda spoils the elegant simplicity of grouping based on filepath. I've taken a crack at the problem but it seems clunky and may have issues. I'd be happy if you've got a better idea.

dnotes:
Well, now that I see the commit, it doesn't look that bad. I think there may still be cases where this would fail, like if someone tried to use a command `behat/behat/sub/behat/behat`, but I would be comfortable not supporting that kind of thing.

achekulaev:
Well, let's give it a shot. Thanks for contribution!

PR #1017, opened 2019-03-29
Windows installation docs update

lmakarov:
- Removed Babun as an installation option. WSL is not the only option available.
- Updated minimal Windows version requirement to Windows 10


docs/content/getting-started/setup.md
```diff
@@ -9,10 +9,6 @@ aliases:
 
 RAM requirement: 8GB or more.
 
-{{% notice info %}}
```
achekulaev: It was on top because https://get.docksal.io/ refers this page 

https://github.com/docksal/get.docksal.io/blob/master/installer.sh#L44

lmakarov: It was a little too prominent at the very top of the getting started docs.
I updated the link in the installer to point to https://docs.docksal.io/getting-started/setup/#install, so users will land right at that message.

PR #1024, opened 2019-04-09
Add Pantheon stack file to files_to_download list

sergiuteaca:
With fin 1.85.0 when doing a `fin update` it doesn't download the newly added Pantheon stack. So, I added the  `stack-pantheon.yml` to the `files_to_download` list in `update_config_files ()` function


lmakarov:
Just released in [v1.12.1](https://github.com/docksal/docksal/releases/tag/v1.12.1)

Issue #846, opened 2018-11-20
Allow custom command path to be specified


mike-potter:
Currently, Docksal looks within the .docksal/commands (or ~/.docksal/commands) folder for custom command scripts.  Because Docksal has the *awesome* ability to recognize the 

```
#: exec_target = cli
```

comment at the top of a script file, these custom commands can also be executed on other non-Docksal (CI) environments.  For example, we have a "build" command that we also use on CI to run "composer install, config-import, clear-cache, etc".

However, we are not comfortable executing a script within the .docksal/commands folder on CI because we only want to use docksal locally and would be deleting the .docksal folder on CI/Stage/Prod/etc.

It would be super useful if there was a configuration path to tell Docksal where to look for the custom command folder.

As a current workaround, we have our scripts in a top-level /scripts folder.  Then for each script we create a custom command script in .docksal/commands that just executes the corresponding script in the /scripts folder.  If would be nice to instead just tell Docksal to look within /scripts directly.

Any chance of getting this simple config request?  It would hugely simplify our project scripting and help encourage even more use of Docksal locally.

achekulaev:
Docksal looks for custom commands in 2 places*: 
- `<project>/.docksal/commands` 
- `$HOME/.docksal/commands`. 

It seems like the second location should work for your CI needs.

--

\* - actually in 4, since the same applies to addons

mike-potter:
No, these are still project-level scripts, not global.  The $HOME/.docksal folder doesn't exist in our CI environment.  We just have the files that are in the repo.  So right now we have /scripts and /.docksal but there is a move to remove the .docksal folder on CI and above environments since they don't need it.

achekulaev:
Why don't you simply make your `<project>/docksal/commands` a symlink to your scripts folder then?

mike-potter:
We might do that.  However, we *do* have a couple of custom commands that are purely for local dev and docksal specific.  So we have those docksal-specific commands in ``<project>/.docksal/commands`` but have the non-specific commands (like rebuild) in ``<project>/scripts``.  In the CI environments it only needs the stuff in ``scripts``

lmakarov:
@mike-potter how about symlinking?

achekulaev:
@lmakarov fin will not work with symlinks for commands or command folder.

achekulaev:
[...]

PR #1036, opened 2019-04-17
Allow help text for commands located in symlinked directories.

jhedstrom:
This addresses the issue noted here https://github.com/docksal/docksal/issues/846#issuecomment-484269763. It does not completely resolve that, but at least allows help text to be printed.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1536,7 +1536,7 @@ show_help_list_commands_in_help ()
 	local executable="-executable"
 	is_mac && executable="-perm +100"
 
-	local addons_list=$(find "$addons_path" -type f ${executable} 2>/dev/null | sort -n 2>/dev/null)
+	local addons_list=$(find -L "$addons_path" -maxdepth 2 -type f ${executable} 2>/dev/null | sort -n 2>/dev/null)
 	if [[ "$addons_list" != "" ]]; then
 		local scope
 		for cmd_name in ${addons_list}


achekulaev:
The fact that symlinks work is an unintended and unsupported usecase. I would discourage anyone to use symlinks instead in place of command files or folders. 

I don't see a good way to process all edge cases:
- symlink to file is broken, fetching help description will fail, command will not work
- someone creates symlink to directory (since [commands can be grouped](https://docs.docksal.io/fin/custom-commands/#grouping-commands)) but that will not work properly
- `exec_target=cli` will not work since file is not in the project directory, but users will blame Docksal
- `find -L` can fall into the endless rabbit's hole if someone creates a wrong link (say to `/etc`)

jhedstrom:
Hmm, so the use case for symlinked directories (or a simple way to configure the location of commands) is outlined in #846. I don't anticipate symlinking these from _outside_ of a project.

As things are now, all commands are found for execution in a symlinked directory, but no help text is displayed.

I'd be happy to work on the configurable command locations if that approach is more solid than supporting symlinks (which are already half supported.)

achekulaev:
@jhedstrom ok I see. I don't really want to open the can of "configurable commands directories" can of worms. Let's stick with allowing symlinks. 

If you add `-maxdepth 2` to that `find` it should resolve the most severe concerns I have about following symlinks. 

And maybe some minor error processing in case of broken symlinks.

jhedstrom:
@achekulaev awesome, thanks!

Regarding error handling, in local testing if I symlink the `.docksal/commands` directory somewhere, and then move the destination (breaking the symlink) the behavior goes back to the current behavior of just showing no custom commands.

achekulaev:
@jhedstrom I rather mean when you have `.docksal/commands/something` linked to a nonexistent file.

bin/fin
```diff
@@ -1536,8 +1536,8 @@ show_help_list_commands_in_help ()
 	local executable="-executable"
 	is_mac && executable="-perm +100"
 
-	local addons_list=$(find "$addons_path" -type f ${executable} 2>/dev/null | sort -n 2>/dev/null)
-	if [[ "$addons_list" != "" ]]; then
+	local addons_list=$(find -L "$addons_path" -maxdepth 2 -type f ${executable} 2>/dev/null | sort -n 2>/dev/null)
+  if [[ "$addons_list" != "" ]]; then
```
achekulaev: please revert breaking the indentation in this line. all indentation is currently tabs

```suggestion
	if [[ "$addons_list" != "" ]]; then
```

jhedstrom:
I tested with a missing file, and the `find` command appears to not find it at all :)

Screen Shot 2019-04-18 at 4 11 45 PM
Screen Shot 2019-04-18 at 4 12 01 PM

Will revert the bad indentation now.

PR #1039, opened 2019-04-22
Instruct docker to clean up exited docksal/empty container

tbtmuse:
Added docker run flag that will cause docker to remove the docksal/empty container after it exits.


achekulaev:
Thanks!

PR #1043, opened 2019-05-01
Reset system after docker machine started

ms-marcus:
I saw on a developers laptop that when he ran `fin up` after a reboot, the resolver for *.docksal was not added and the NFS mounts weren't properly setup. I noticed that when he was doing this docksal was saying that the docker-machine wasn't started and offering to start it for him, however it said nothing about NFS mounts and resolvers being added, whereas `fin project restart` did, and fixed the issue.

This PR adds the missing step to the `docker_machine_start` routine. I'm not convinced this is the correct place to put it yet as `check_docker_running` (and therefore possibly `docker_machine_start`) is called by `system_reset` which feels wrong. However, it shouldn't cause any loops as by the time `docker_machine_start` calls `system_reset` the docker machine should be started and therefore the `docker_machine_start` code is never called. I'm open to suggestions on where this would be better placed :)

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -2424,6 +2424,9 @@ docker_machine_start ()
 			exit 1
 		fi
 	fi
+
+	# After docker machine is restarted, we should reset the system to make sure everything is in a sane state.
+	system_reset
 }
 
 docker_machine_remove ()


lmakarov:
@ms-marcus `fin up` / `fin start` are (respectively) identical to `fin project up` / `fin project start`. There is a marginal difference between `up` and `start` - `up` forces config reload.

I'm not able to reproduce this:

```
$ fin up
It looks like 'docksal' docker machine is not running.
Start it now? [y/n]: y
Starting "docksal"...
(docksal) Check network to re-create if needed...
(docksal) Waiting for an IP...
Machine "docksal" was started.
Waiting for SSH to be available...
Detecting the provisioner...
Started machines may have new IP addresses. You may need to re-run the `docker-machine env` command.
 ALERT:  Some Docksal services were not running
         Restarting Docksal system services...
Configuring network settings...
Configuring NFS shares...
Writing /etc/exports...
Restarting nfsd...
Mounting NFS shares...
Mounting local /Users/leonid to /Users/leonid
Resetting Docksal system services...
 * proxy
 * dns
   upstream 192.168.0.1
Enabling automatic *.docksal DNS resolver...
Clearing DNS cache...
 * ssh-agent
Identity added: id_ecdsa (id_ecdsa)
Starting services...
...
```

Please check that the latest Docksal version is used, as there was major refactoring in the recent [v1.12.0](https://github.com/docksal/docksal/releases/tag/v1.12.0) release related to the bootstrap and update functionality in `fin`. If this is reproducible on an up-to-date system, then please file a bug report with the necessary details.

PR #1047, opened 2019-05-02
Fixed issue where ngrok cannot connect to a container if other argument

### LINKED ISSUE #1044: fin share with --host-header doesn't seem to work
**Description**

fin version 1.85.1

Running `fin share --host-header=special.baxter.docksal` results in an error:

```
ERROR:  You must specify a single argument: a port or address to tunnel to.
ERROR:  You specified 0 arguments: []
ERROR:  For example, to expose port 80, run 'ngrok http 80'.
```

If I insert `echo $ARGS` after the docker run command in ngrok_share, I get the following output after the command fails:

```
ngrok http -host-header=special.baxter.docksal
```

Without the addition of `--host-header=special.baxter.docksal` the following is output after the command:

```
ngrok http -host-header 'baxter.docksal' corvette_web_1.corvette_default:80
```

So, when `--host-header` is specified it looks like corvette_web_1.corvette_default:80 needs to be attached. If I run `fin share --host-header='special.baxter.docksal corvette_web_1.corvette_default:80'` then the whole thing works, but I don't think we should have to do that.

sean-e-dietrich:
Fixes #1044 

This also makes it so if there is a different container other than web

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -5466,9 +5466,11 @@ ngrok_share ()
 	eval $(parse_params "$@")
 
 	local network="${COMPOSE_PROJECT_NAME_SAFE}_default"
-	local container_name=${container:-$(docker-compose ps web | grep Up | grep _web_ | cut -d " " -f 1)}
+	local container=${container:-web}
+	local container_name=$(docker-compose ps ${container} | grep Up | grep _${container}_ | cut -d " " -f 1)
+
 	if [[ "$container_name" == "" ]]; then
-		echo-error "Could not find running web container in this project"
+		echo-error "Could not find running ${container} container in this project"
 		exit 1
 	fi
 	local ngrok_container_name=${container_name}_ngrok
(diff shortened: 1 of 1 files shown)


achekulaev:
@sean-e-dietrich so that $container can be set like an ENV var? Should it be all caps then?

sean-e-dietrich:
@achekulaev i thought about that. Right now we have all lower case flags that are being used with Ngrok but we also allow for uppercase variables to be used. This is the one variable we just modified. So yes, we should allow someone to set it via a variable.

achekulaev:
Ah it's a flag, I see. Thanks

bin/fin
```diff
@@ -5567,7 +5569,12 @@ ngrok_share ()
 
 	# Add -host-header by default if no arguments are included.
 	if [[ "${ARGS}" == "ngrok http" ]]; then
-		ARGS="${ARGS} -host-header '${NGROK_VIRTUAL_HOST}' ${container_name}.${network}:${NGROK_PORT}"
+		ARGS="${ARGS} -host-header '${NGROK_VIRTUAL_HOST}'"
+	fi
+
+	# If no configuration was used we append main container.
```
achekulaev: Could you please explain better why we only append main container if CONF_ARGS is empty.
I am asking because I genuinely have no idea why myself :)

sean-e-dietrich: When you use configuration files you can't also specify the container. It's an either or situation. So, if there is no configuration files we append the container on the end.

achekulaev: Why? Because its ngrok limitation, or because of logical conflict?

sean-e-dietrich: @achekulaev this is an ngrok limitation. So far I have found that it uses either the config files or it uses the command line arguments.

lmakarov:
Needs to be rebased onto develop and tests updated/fixes (if necessary).

achekulaev:
I don't see any conflicts so rebasing doesn't look like a required step. Approving

lmakarov:
Can we just have this done as part of #1048 and ensure tests are green?

PR #1048, opened 2019-05-02
Ngrok Modification

### LINKED ISSUE #1046: Rework fin share as part of project
Docker is deprecating the `--link` argument and is recommended to create specific networks and add that container to the network.

Find out way to implement Ngrok on a project level

sean-e-dietrich:
Modified Ngrok to connect to a docker project network.

This is also helpful when utilizing the configuration files to reference the hostname of the service.

Fixes #1046

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -5466,9 +5466,11 @@ ngrok_share ()
 	eval $(parse_params "$@")
 
 	local network="${COMPOSE_PROJECT_NAME_SAFE}_default"
-	local container_name=${container:-$(docker-compose ps web | grep Up | grep _web_ | cut -d " " -f 1)}
+	local container=${container:-web}
+	local container_name=$(docker-compose ps ${container} | grep Up | grep _${container}_ | cut -d " " -f 1)
+
 	if [[ "$container_name" == "" ]]; then
-		echo-error "Could not find running web container in this project"
+		echo-error "Could not find running ${container} container in this project"
 		exit 1
 	fi
 	local ngrok_container_name=${container_name}_ngrok
@@ -5567,7 +5569,12 @@ ngrok_share ()
 
 	# Add -host-header by default if no arguments are included.
 	if [[ "${ARGS}" == "ngrok http" ]]; then
-		ARGS="${ARGS} -host-header '${NGROK_VIRTUAL_HOST}' ${container_name}.${network}:${NGROK_PORT}"
+		ARGS="${ARGS} -host-header '${NGROK_VIRTUAL_HOST}'"
+	fi
+
+	# If no configuration was used we append main container.
+	if [[ -z $CONF_ARGS ]]; then
+		ARGS="${ARGS} ${container_name}.${network}:${NGROK_PORT}"
 	fi
 
 	${winpty} docker run --rm ${NGROK_VOLUME} -it \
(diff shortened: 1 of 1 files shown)


achekulaev:
Looks great to me. I would keep it open until 0.12.2 releases.

lmakarov:
@sean-e-dietrich please fix tests.

sean-e-dietrich:
@lmakarov this has been updated and retested.

bin/fin
```diff
@@ -8000,6 +8011,8 @@ case "$1" in
 		[[ "$USED_ALIAS" != "" ]] &&
 			[[ "$(pwd)" == "$PROJECT_ROOT" ]] &&
 			cd "$DOCROOT"
+                [[ "${NO_DRUSH_OPTIONS_URI}" != "" ]] &&
```
lmakarov: @sean-e-dietrich does this change belong here?

lmakarov:
@sean-e-dietrich please see my thoughts in https://github.com/docksal/docksal/issues/1046#issuecomment-539239430 regarding an alternative approach.

lmakarov:
@achekulaev The alternative approach I'm suggesting is doing the same thing, but with docker-compose, instead of bare docker. The end result is the same, but with less code and more reliable.

>  too many open questions

What are the open questions?

achekulaev:
@lmakarov 

Open questions of the alternate approach are:

- How to get ngrok URL after I started sharing?
- How to pass all additional ngrok params, that are currently passed as command line arguments?
- ngrok used to be a temporary share thingy that now converts into a permanent feature that has ti be turned off, it should be highlighted that it will continue sharing until you cancel, because it is a security implication
- Someone needs to actually Test new approach
- Someone needs to create new docs on using the new approach, that will include answers to all the questions above

I estimate the effort size to address these questions to be medium-to-large. Consequently to avoid pushing 1.13 release date further I believe it will be better to put it off till 1.14.

sean-e-dietrich:
@achekulaev 

> How to get ngrok URL after I started sharing?
We will output the URL to the person using the Ngrok API

> How to pass all additional ngrok params, that are currently passed as command line arguments Can they be passed as ENV vars or does it need to be implemented additionally?
We previously allowed for ENV vars and they were just mapping the existing container variables. Additionally, we allowed for the command line args to override the environment variables. This is still going to be the case.

> ngrok used to be a temporary share thingy, that turns off quickly by ctrl-c. With container approach it would have to be turned off to avoid sharing, which should be highlighted in the docs, because it might be a security implication
We can have little messages run every so often that share is still running so they are aware of it.

> Someone needs to actually Test new approach
We will need to adjust existing tests to use new approach but just a minor change to look at new service for the new container. Additionally, there will need to be more tests added.

> Someone needs to create new docs on using the new approach, that will include answers to all the questions above
This is the same with everything we do.

Take a look here as a start and maybe this will help answer some questions

https://github.com/docksal/docksal/pull/1195

achekulaev:
@sean-e-dietrich I used "open questions" meaning matters that require time investment to get them ironed out, I did not mean questions literally. Of course I understand and/or foresee the answers to those questions, but the gist the comment is that there are those things and to avoid pushing 1.13 release date further I believe it will be better to put it off till 1.14. 

But if you guys feel like getting it done in 1.13 I do not want to hold the enthusiasm in any way. We just need it implemented ,tested and documented. I am totally ok to review a final PR from whoever does that.

PR #1057, opened 2019-05-07
Export CLI user to a configurable variable

rhabbachi:
Currently the user used to execute command on the `cli` container is hard coded to `docker`. This works well for the "official" docksal `cli` container provided in the stacks but may not be true for custom `cli` container.

This changes exports the `cli` user to a `CLI_USER` variable that can be set via the various docksal environments overrides.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -306,6 +306,8 @@ IMAGE_SSH_AGENT=${IMAGE_SSH_AGENT:-docksal/ssh-agent:1.2}
 IMAGE_VHOST_PROXY=${IMAGE_VHOST_PROXY:-docksal/vhost-proxy:1.5}
 IMAGE_DNS=${IMAGE_DNS:-docksal/dns:1.1}
 
+# User to be passed to the CLI container when executing commands.
+CLI_USER=${CLI_USER:-docker}
 #---------------------------- Helper functions --------------------------------
 
 DOCKSAL_PATH='' #docksal path value will be cached here
@@ -4996,7 +4998,7 @@ _exec ()
 
 	if [[ "$container_name" == "cli" ]]; then
 		# Commands in cli should be run using the docker user, not root.
-		local container_user='-u docker'
+		local container_user="-u $CLI_USER"
 	fi
 
 	# Enter project containers
--- a/docs/content/stack/configuration-variables.md
+++ b/docs/content/stack/configuration-variables.md
@@ -209,6 +209,12 @@ Acquia Cloud API Key. See [Acquia Drush Commands](/tools/acquia-drush/).
 
 Token used for logging in to Pantheon's CLI Tool [Terminus](/tools/terminus/).
 
+### CLI_USER
+
+`Default:  docker`
+
+User to be used for `fin exec`.
+
 ## Image Variables
 
 The following variables are used to set images to a specified version. Using them will prevent the images from being udpated


achekulaev:
@rhabbachi looks like a reasonable change to me. 

Some notes.

1. Since it was added to `_exec` it should probably also be added to `run_cli` 
1. Could you please add the description of this variable to the appropriate documentation? https://github.com/docksal/docksal/blob/develop/docs/content/stack/configuration-variables.md

rhabbachi:
Hey @achekulaev,
I missed the `run_cli` use case, good catch. I've been looking into it, but seems like it's more problematic to reproduce the same behavior. `run_cli` does not load the global/project config before hand and does not have access to the CLI_USER variable (among other things) https://github.com/docksal/docksal/blob/develop/bin/fin#L7814

I haven't used the `run-cli` command much and looking at the code it seems to me it's a bit of an "outsider" (does not use the current IMAGE_CLI to run commands, no access to various project/global environment vars/config, etc) so not sure how to proceed for it.

rhabbachi:
No updates?

lmakarov:
> This works well for the "official" docksal cli container provided in the stacks but may not be true for custom cli container.

@rhabbachi can you please elaborate on your use case. What is the custom cli container that you are using?

`cli` is a unique service in Docksal. `fin` makes a lot of assumptions based on how `cli` works. I would not recommend replacing the stock one with a completely custom one, unless you really know what you are doing.

rhabbachi:
@lmakarov Hey!
I'm using wodby's [drupal-php](https://github.com/wodby/drupal-php) as the cli container. The username there is wodby, hence the `CLI_USER` variable.

I built and currently supporting multiple deployments that uses the wodby stack + additional custom services, and currently converted one of those from [a custom docker-compose wrapper](https://github.com/rhabbachi/robocomp-dkan) to docksal. The idea being to keep the stack and patch `docksal` to support it.

I've been looking at docksal's `bin/fin` script. The thing is, docksal is basically what I want from a docker/docker-compose single node multi-env deployment tool: bash so deploys everywhere without a hassle/dependencies, multiple environments/overrides via `.docksal/docksal-<env>.yml`, custom commands... Obviously as you mentioned some parts are not designed for non local deployments nor custom services (This PR, but also the system stack is basically hard coded and had to workaround it cf your [poc](https://github.com/lmakarov/docksal-prod-poc)). If the vision for docksal include it being generic enough to support other use cases, maybe I can help with some PRs.

lmakarov:
@rhabbachi fair enough. As adding `CLI_USER` does not hurt anything, I don't have objections.

> I built and currently supporting multiple deployments that uses the wodby stack + additional custom services, and currently converted one of those from a custom docker-compose wrapper to docksal

Would you be able to share your custom setup? It would be a nice show case of flexibility built into Docksal/fin.

> Since it was added to `_exec` it should probably also be added to `run_cli`

@achekulaev I think we can skip `run_cli` here. It's a very unique use case that fully relies on using the stock `docksal/cli` image.

Let's keep this PR open until 1.12.2 is released.

rhabbachi:
@lmakarov I've been dumping the substitue system stack I'm using to [this repository](https://github.com/rhabbachi/notlocal). Obviously inspired from your poc.

I'm coming up on my current client contract and may have more time to push the wodby stack I've been using for the drupal project soon.

1.12.**3** was released recently, should this PR be merged?

lmakarov:
@rhabbachi yep, merged!

Thanks for sharing your work in https://github.com/rhabbachi/notlocal. Looks like this would be another case for a global addon/plugin. CC @achekulaev

lmakarov:
Heads up - this PR merge will be reverted and replaced with #1122

rhabbachi:
@lmakarov Thanks for the heads up! Changes done in #1122 are reallly sweet!

This maybe not really the best place to do it, but quick feedback: I'm in the process of leveraging docksal awesomeness to manage local and "hosted" envs (including production). I have couple changes done that I'm planning to contribute back. Specifically we use similar docksal stack for two distinct frontend and backend repositories for local. The same stack is deployed for MR (gitlab oblige) env + stage and production. non local hosts are managed by custom global addon based off of the notlocal addon linked above that deploys app stacks, CI registry logins, and any handy script to manage a docksal host.

Really thankful for your work on this project!

lmakarov:
@rhabbachi if you want, feel free to write a blog post about your setup on Medium. We can then include it as a guest post on blog.docksal.io :)

rhabbachi:
Great idea, will definitely ping when I get around to it!

PR #1066, opened 2019-05-28
Update fin-run-cli.md

frankfarm:
Typo


lmakarov:
Yeah, fin and fun are almost the same thing ;) Thanks!

PR #1080, opened 2019-06-13
Adds SSH key forwarding support for cli

### LINKED ISSUE #1079: Add support for SSH key forwarding in CI
There are a few ways how SSH keys can be passed to an environment on a Docksal Sandbox server:

1. Store the keys on the server in the `build-agent` user's directory (`/home/build-agent/.ssh`) and load then into the `docksal-ssh-agent` service. This works the same way as with a local Docksal instance.

2. Pass as a [secret](https://github.com/docksal/service-cli#secrets-and-integrations) in `SECRET_SSH_PRIVATE_KEY `. Secrets are [written](https://github.com/docksal/ci-agent/blob/82623a009a8d79d7e6d54c8804f083673acf9fa5/base/bin/build-init#L90-L101) into the project's `docksal-local.env` file by `docksal/ci-agent`. `cli` supports reading the private SSH key from this variable.

When working in a shared environment (many projects on the same sandbox server), both approaches may be insecure.

(1) gives any project (`cli` container) on the server access to all keys loaded in the agent
(2) exposes private keys in plain text in project's code base on the server

To better secure private SSH keys we can use SSH key forwarding, which can be chained:

```
ci-agent (keys loaded in ssh-agent) = SSH => server = (mounted SSH_AUTH_SOCK) => cli container
```

lmakarov:
Fixes #1079

SSH key forwarding chain:

```
ci-agent (has keys loaded in ssh-gent) 
  => server (SSH_AUTH_SOCK) 
      => project cli container (mounted SSH_AUTH_SOCK from the host and overridden SSH_AUTH_SOCK variable)
```

This PR implements the last link of the chain.

SSH key(s) loaded in the ssh-agent on the client will be available on the server and in the target project's `cli` container only during the build time. Once the SSH connection from client drops, the ssh-agent socket becomes unavailable.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -253,6 +253,15 @@ PATH="$CONFIG_BIN_DIR:$PATH"
 # Set global variable in case native Docker app is used/not-used
 DOCKER_NATIVE="${DOCKER_NATIVE:-0}"
 
+# SSH agent settings
+if ! is_ci; then
+	# Use the built-in agent in non-CI environments
+	DOCKSAL_SSH_AGENT_USE_HOST="${DOCKSAL_SSH_AGENT_USE_HOST:-0}"
+else
+	# Prefer the hosts' agent in CI environments
+	DOCKSAL_SSH_AGENT_USE_HOST="${DOCKSAL_SSH_AGENT_USE_HOST:-1}"
+fi
+
 #---------------------------- URL references --------------------------------
 GITHUB_API="https://api.github.com"
 URL_REPO="https://raw.githubusercontent.com/docksal/docksal"
@@ -7519,9 +7528,10 @@ else
 	export DOCKER_RUNNING="false"
 fi
 
-# Only allow SSH key forwarding in a CI/Linux environment, otherwise cli won't be able to start on Mac/Win,
-# since ssh-agent sockets there would be outside of the mounted/accessible projects directory.
-if ! is_ci || ! is_linux; then unset SSH_AUTH_SOCK; fi
+# Reusing host's ssh-agent socket (SSH_AUTH_SOCK) is only supported on Linux.
+# On Mac/Win, SSH_AUTH_SOCK would be outside of the mounted projects directory and thus inaccessible (cli will fail to start).
+# On Linux, we default to the built-in agent, unless DOCKSAL_SSH_AGENT_USE_HOST=1
+if ! is_linux || [[ "${DOCKSAL_SSH_AGENT_USE_HOST}" != "1" ]]; then unset SSH_AUTH_SOCK; fi
 
 # Handle Alias
 if [[ "$1" == "@"* ]]; then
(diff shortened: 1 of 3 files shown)


achekulaev:
1. Lacks documentation on CI and SSH_AUTH_SOCK connection
2. ssh-key functions should throw alerts/errors if CI is enabled since they become obsolete without ssh agent volume mapped to cli

bin/fin
```diff
@@ -7519,6 +7519,10 @@ else
 	export DOCKER_RUNNING="false"
 fi
 
+# Only allow SSH key forwarding in CI (Linux), otherwise cli won't be able to start on Mac/Win, since ssh-agent
+# sockets there would be outside of the mounted/accessible projects directory.
+if ! is_ci; then unset SSH_AUTH_SOCK; fi
```
```suggestion
if ! is_ci || ! is_linux; then unset SSH_AUTH_SOCK; fi
```

lmakarov: Updated

lmakarov:
> Lacks documentation

@achekulaev for now, this will be documented in https://github.com/docksal/ci-agent, until we have a section on CI and Docksal Sandbox Server on docs.docksal.io.

docs/content/stack/configuration-variables.md
```diff
@@ -30,6 +30,20 @@ Docker image to use as the VHOST Proxy. This forwards all requests to the approp
 
 Docker image to use for DNS Routing.
 
+### CI
+
+This is a global variable. When set, enables CI mode:
+
+- Sets `DOCKSAL_VHOST_PROXY_IP="0.0.0.0"`, which opens vhost-proxy to the world.
+- Reports the instance as a CI instance (vs local), so our anonymous usage stats are not skewed.
+- Allows SSH agent forwarding from ci/dev to sandbox server to cli containers (only works on Linux servers).
```
achekulaev:
```suggestion
- Allows [SSH agent forwarding](https://developer.github.com/v3/guides/using-ssh-agent-forwarding/) that makes SSH keys which were set on the CI server to be automatically available inside `cli` containers by automatically passing `SSH_AUTH_SOCK` variable to them (only works on Linux servers). 
```

frederickjh:
> > Lacks documentation
> 
> @achekulaev for now, this will be documented in https://github.com/docksal/ci-agent, until we have a section on CI and Docksal Sandbox Server on docs.docksal.io.

I do not use either of those and still use ssh agent forwarding to do Drupal deployments with [Deployer](https://deployer.org).

lmakarov:
@achekulaev I decided to add `DOCKSAL_SSH_AGENT_USE_HOST` variable, which controls which ssh-agent socket will be used by default (local vs CI) and allows user overrides (on Linux only).

Eventually, we can figure something out on Mac and Win to allow users to reuse their host's ssh-agent.

@GMTA FYI

PR #1081, opened 2019-06-13
Allow disabling checking for updates

GMTA:
This change adds a new environment variable `DOCKSAL_CHECK_FOR_UPDATES` which allows disabling automatically checking for updates. The default value is `true` but can be overridden by setting an environment variable or setting a value in `docksal.env`.

Rationale: we automatically deploy Docksal for development purposes and test new versions before releasing them across the board. If Docksal always automatically checks for new updates, users will be encouraged to run `fin update` instead of running on the fixed version we deploy. Using this new configuration value, we can disable the automatic update checks preventing the "fin update" message.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -247,8 +247,8 @@ DOCKSAL_STATS_TID='UA-93724315-1'
 DOCKSAL_STATS_URL='http://www.google-analytics.com/collect'
 DOCKSAL_STATS_OPTOUT=${DOCKSAL_STATS_OPTOUT:-0}
 
-# Whether to disable the automatic check for Docksal updates
-DOCKSAL_DISABLE_UPDATES="${DOCKSAL_DISABLE_UPDATES:-0}"
+# Whether to prevent (checking for) updates for Docksal
+DOCKSAL_LOCK_UPDATES="${DOCKSAL_LOCK_UPDATES:-0}"
 
 # Override PATH to use our utilities
 PATH="$CONFIG_BIN_DIR:$PATH"
@@ -4863,7 +4863,7 @@ check_for_updates ()
 		echo "$timestamp" > "$CONFIG_LAST_PING"
 	fi
 
-	[[ "$DOCKSAL_DISABLE_UPDATES" == "1" ]] && return # Do not check for updates if disabled
+	[[ "$DOCKSAL_LOCK_UPDATES" == "1" ]] && return # Do not check for updates if disabled
 
 	if wants_rc_version; then
 		# Check once a day if DOCKSAL_USE_RC is set
@@ -7771,7 +7771,7 @@ case "$1" in
 		vm "$@"
 		;;
 	update)
-		if [[ "$DOCKSAL_LOCK_UPDATES" != "" ]]; then
+		if [[ "$DOCKSAL_LOCK_UPDATES" == "1" ]]; then
 			echo-error "Updates locked in this environment"
 			exit 1
 		fi
--- a/bin/fin
+++ b/bin/fin
@@ -4842,10 +4842,6 @@ get_fin_url ()
 
 check_for_updates ()
 {
-	if [ "${DOCKSAL_CHECK_FOR_UPDATES}" != 'true' ]; then
-		return
-	fi
-
 	# Never trigger in scripts
 	if ! is_tty; then return; fi
 	local UPDATE_AVAILABLE=0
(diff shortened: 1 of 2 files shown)


achekulaev:
`check_for_updates` function also helps us collect usage statistics, by sending anonymous pings to Google Analytics. Please move your check a little below, after the `stats_ping`.

GMTA:
@achekulaev Good point, I missed that one. Moved the return a bit down!

lmakarov:
@GMTA instead (or in addition) you can pin the Docksal version.
Even if someone runs `fin update`, they will get a version you did not approve internally.

```
$ fin config set --global DOCKSAL_VERSION=v1.12.2
Replaced value for DOCKSAL_VERSION in /home/build-agent/.docksal/docksal.env
$ fin update
[!] Using Docksal version: v1.12.2
[STEP 1/4] Updating fin...
fin 1.86.0 downloaded...
...
```

```
$ fin config set --global DOCKSAL_VERSION=v1.12.3
Replaced value for DOCKSAL_VERSION in /home/build-agent/.docksal/docksal.env
$ fin update
[!] Using Docksal version: v1.12.3
[STEP 1/4] Updating fin...
fin 1.86.2 downloaded...
...
```

bin/fin
```diff
@@ -247,6 +247,9 @@ DOCKSAL_STATS_TID='UA-93724315-1'
 DOCKSAL_STATS_URL='http://www.google-analytics.com/collect'
 DOCKSAL_STATS_OPTOUT=${DOCKSAL_STATS_OPTOUT:-0}
 
+# Whether to automatically check for Docksal updates
+DOCKSAL_CHECK_FOR_UPDATES="${DOCKSAL_CHECK_FOR_UPDATES:-true}"
```
```suggestion
DOCKSAL_DISABLE_UPDATES="${DOCKSAL_DISABLE_UPDATES:-0}"
```

bin/fin
```diff
@@ -4860,6 +4863,8 @@ check_for_updates ()
 		echo "$timestamp" > "$CONFIG_LAST_PING"
 	fi
 
+	[[ "$DOCKSAL_CHECK_FOR_UPDATES" != "true" ]] && return # Do not check for updates if disabled
```
```suggestion
	[[ "$DOCKSAL_DISABLE_UPDATES" == "1" ]] && return # Do not check for updates if disabled
```

bin/fin
```diff
@@ -247,6 +247,9 @@ DOCKSAL_STATS_TID='UA-93724315-1'
 DOCKSAL_STATS_URL='http://www.google-analytics.com/collect'
 DOCKSAL_STATS_OPTOUT=${DOCKSAL_STATS_OPTOUT:-0}
 
+# Whether to automatically check for Docksal updates
```
```suggestion
# Whether disable automatic check for Docksal updates
```

achekulaev:
@GMTA I propose the other way around approach in naming to highlight the default, which is checking for updates.

GMTA:
@achekulaev Good suggestion; change has been pushed

GMTA:
@lmakarov 

> you can pin the Docksal version

Good suggestion! Going to use this for now.

lmakarov:
@GMTA assuming pinning a version with `DOCKSAL_VERSION` works for you, is there still a solid need in `DOCKSAL_DISABLE_UPDATES`?

GMTA:
@lmakarov Disabling the updates could also increase the performance of `fin` by not having to check for updates and the last check timestamp file. However, this is currently not part of this PR since the timestamp file is read regardless of the setting.

Apart from that, having the option to disable the automatic updates comes in useful when testing different versions of Docksal without having to update `DOCKSAL_VERSION` every time to prevent the checks and accompanying message.

lmakarov:
Please document the new variable in `docs/content/stack/configuration-variables.md`.

GMTA:
@lmakarov Documentation updated

lmakarov:
@GMTA reviewing the docs I see we already have a variable that prevents updates, but not necessarily update checks (what you are asking for) - `DOCKSAL_LOCK_UPDATES`

https://github.com/docksal/docksal/blob/c5afbf47d20d5d5017ebdac9a71d63ef7f08a741/bin/fin#L7769

Let's use the existing variable and extend it to also disable update checks.

The check above will need to be updated since right now `DOCKSAL_LOCK_UPDATES=0` won't work.

Proposed change:

```
if [[ "$DOCKSAL_LOCK_UPDATES" == "1" ]]; then
```

Also docs updated accordingly.

We are almost there :)

GMTA:
@lmakarov Good find; updated!

PR #1088, opened 2019-06-24
Implement an SSH auth proxy over TCP for Mac / Windows

GMTA:
This implements an SSH proxy in Docksal to support SSH agent forwarding for Mac / Windows. This means that key management through Docksal is now optional and you can use your local SSH agent such as the built-in `ssh-agent`, KeePass or really anything that sets `SSH_AUTH_SOCK`.

To make this work, the following MR in `service-ssh-agent` is needed:
https://github.com/docksal/service-ssh-agent/pull/8

**Usage**
- Set `DOCKSAL_SSH_AGENT_USE_HOST` to `1` (no longer limited to Linux)
- Run `fin system reset ssh-agent` to make sure the local proxy is started
- Run `fin up` within a project to make use of the SSH proxy

**Implementation details**
- Uses `socat` to start a local proxy. Unfortunately, this tool is not available by default on Mac or Windows. However, it is easily installed by `brew install socat` or `apt-get install socat` which is probably possible if you use Docksal on those platforms. The user receives instructions when `socat` is unavailable.
- Exposes a TCP socket on `DOCKSAL_HOST_IP` on port `30001` by default. The port was chosen at random. This also implies a security risk: anyone with access to `DOCKSAL_HOST_IP` (which is usually an internal IP address unreachable by outsiders) can use this port to access your local SSH agent. However, SSH proxying is disabled by default since `DOCKSAL_SSH_AGENT_USE_HOST` is `0` if you are not running Docksal in a CI environment. Ultimately, I think that the responsibility lies with the (well-informed) developer to choose between usability and security.
- The TCP port can be configured through the `DOCKSAL_SSH_PROXY_PORT` environment variable.
- Disables `fin ssh-key ...` commands if running with `DOCKSAL_SSH_AGENT_USE_HOST=1` since those commands will not work.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -7593,7 +7593,7 @@ fi
 if [[ -n "$SSH_AUTH_SOCK" ]]; then
 	DOCKSAL_SSH_AUTH_SOCK="$SSH_AUTH_SOCK"
 fi
-if ! is_linux || [[ "${DOCKSAL_SSH_AGENT_USE_HOST}" != "1" ]]; then
+if ! is_linux || [[ "$DOCKSAL_SSH_AGENT_USE_HOST" != "1" ]]; then
 	unset SSH_AUTH_SOCK
 fi
 
--- a/bin/fin
+++ b/bin/fin
@@ -4056,12 +4056,11 @@ install_sshagent_service ()
 
 	# Switch between built-in agent and auth socket proxying
 	local ssh_service_args
+	stop_sshproxy_daemon
 	if is_proxy_sshagent; then
-		stop_sshproxy_daemon
 		start_sshproxy_daemon
 		ssh_service_args="ssh-proxy $DOCKSAL_HOST_IP $DOCKSAL_SSH_PROXY_PORT"
 	else
-		stop_sshproxy_daemon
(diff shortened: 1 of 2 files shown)


bin/fin
```diff
@@ -4045,14 +4054,57 @@ install_sshagent_service ()
...
+	if is_proxy_sshagent; then
+		start_sshproxy_daemon
+		ssh_service_args="ssh-proxy $DOCKSAL_HOST_IP $DOCKSAL_SSH_PROXY_PORT"
+	else
+		ssh_service_args="ssh-agent"
+	fi
+
 	docker volume create --name docksal_ssh_agent >/dev/null 2>&1
 	docker run -d --name docksal-ssh-agent --label "io.docksal.group=system" --restart=unless-stopped \
+		-e DOCKSAL_SSH_AGENT_USE_HOST \
```
```suggestion
```

No need in passing this.
See my suggestion in https://github.com/docksal/service-ssh-agent/pull/8#issuecomment-519291243

GMTA:
Updated https://github.com/docksal/service-ssh-agent/pull/8 and committed your suggested change here.

PR #1108, opened 2019-08-04
Add ability to list containers that should use bash instead of sh.

### LINKED ISSUE #1107: Allow containers other than `cli` to use bash by default
**Description**

When trying to set up a docksal instance to simulate a decoupled site using different servers (Default stack with additional CLI for React/Node/Gatsby type) I noticed that in order to run commands like `fin exec --in=nodeapp yarn start` or `fin exec --in=nodeapp npm install` where `nodeapp` is the name of the service, I had to go through some significant hoops to get it to work.

1. Add a second CLI based service to `docksal.yml`
2. Try to run a command in the service `fin exec --in=nodeapp npm install`

`sh: 1: npm: not found`

Things to install

JDDoesDev:
Fixes #1107 

Adds the ability for developers to choose which containers use bash and the docker user instead of limiting to `cli` for bash and -u docker and `db` for bash.

To use, add an environment variable to your docksal.env file `CLI_CONTAINERS`

This variable should create a string of container names, separated by a single space.  Within the `fin` script, this string is converted to an array and current container names are compared with the containers listed so that commands run will use the selected shell, bash or sh.

For example, with a container named `nodeapp` for a front-end of a decoupled drupal install the commands before to run NPM functions would be
`container_user="-u docker" fin exec --in=nodeapp npm -v`

or to use `fin bash nodeapp` you would have to get into the container, then run `bash` to change shells, and then `su docker` to change to the docker user to use these commands.  This also meant that to exit, you would have to type `exit` multiple times to get out of the container.

Now, with this PR, you can select which containers should be bash and docker by default with the environment variable `CLI_CONTAINERS`

For example:
`CLI_CONTAINERS="nodeapp otherservice"`

Run `fin p start` to reload the variables, and you can run `fin bash nodeapp` and be in the container as the docker user, or run `fin exec --in=nodeapp npm -v` and not get an error from sh.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -322,7 +322,8 @@ IMAGE_DNS=${IMAGE_DNS:-docksal/dns:1.1}
 CLI_USER=${CLI_USER:-docker}
 
 # Shell to use for CLI commands
-FIN_SHELL=${FIN_SHELL:-sh}
+cli_containers="$CLI_CONTAINERS"
+IFS=' ' read -a bash_cli_containers <<< "$cli_containers"
 #---------------------------- Helper functions --------------------------------
 
 DOCKSAL_PATH='' #docksal path value will be cached here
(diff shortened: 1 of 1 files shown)


achekulaev:
@JDDoesDev while this solution is simplistic, it is somewhat artificial. It would really be much more flexible to use Docker labels, i.e. one defines a label in your `docker.yml` for a container something like

```yaml
  mycontainer:
    ...
    labels:
      - io.docksal.shell=bash
```

Then you would look for the value of this label with 

```bash
docker ps <container> --format '{{.Label "io.docksal.shell"}}'
``` 

and act accordingly. 

This would actually allow to declare the use of `zsh` or other shell.

JDDoesDev:
Thank you @achekulaev.  @lmakarov mentioned similar in the issue thread.  i'll work on that a little bit later and resubmit.

lmakarov:
Closing. Please submit a new PR using the suggested implementation approach.

lmakarov:
Replaced with https://github.com/docksal/docksal/pull/1122

PR #1115, opened 2019-08-08
Adds DRUSH_OPTIONS_URI to the cli service.

### LINKED ISSUE #1114: Add DRUSH_OPTIONS_URI to cli container
**Description**
Setting `DRUSH_OPTIONS_URI` to `http://${VIRTUAL_HOST}` will allow projects using Drush 9 to have the `drush uli` command output a usable URL rather than the current `http://default/...`. Other things like indexing search content, etc, via drush will also then have the proper url.

PR to come.

jhedstrom:
- Closes #1114

### DIFF
--- a/stacks/services.yml
+++ b/stacks/services.yml
@@ -147,7 +147,7 @@ services:
       - XDEBUG_CONFIG=remote_connect_back=0 remote_host=${DOCKSAL_HOST_IP}  # Point xdebug to the host IP
       - PHP_IDE_CONFIG=serverName=${VIRTUAL_HOST}
       - DRUSH_ALLOW_XDEBUG=${XDEBUG_ENABLED:-0}
-      - DRUSH_OPTIONS_URI=${DRUSH_OPTIONS_URI:-http://${VIRTUAL_HOST}}
+      - DRUSH_OPTIONS_URI=${VIRTUAL_HOST}
       - BLACKFIRE_CLIENT_ID
       - BLACKFIRE_CLIENT_TOKEN
       - SECRET_SSH_PRIVATE_KEY
--- a/stacks/services.yml
+++ b/stacks/services.yml
@@ -147,6 +147,7 @@ services:
       - XDEBUG_CONFIG=remote_connect_back=0 remote_host=${DOCKSAL_HOST_IP}  # Point xdebug to the host IP
       - PHP_IDE_CONFIG=serverName=${VIRTUAL_HOST}
       - DRUSH_ALLOW_XDEBUG=${XDEBUG_ENABLED:-0}
+      - DRUSH_OPTIONS_URI=http://${VIRTUAL_HOST}
       - BLACKFIRE_CLIENT_ID
       - BLACKFIRE_CLIENT_TOKEN
       - SECRET_SSH_PRIVATE_KEY
--- a/stacks/services.yml
+++ b/stacks/services.yml
@@ -147,7 +147,7 @@ services:
       - XDEBUG_CONFIG=remote_connect_back=0 remote_host=${DOCKSAL_HOST_IP}  # Point xdebug to the host IP
       - PHP_IDE_CONFIG=serverName=${VIRTUAL_HOST}
       - DRUSH_ALLOW_XDEBUG=${XDEBUG_ENABLED:-0}
-      - DRUSH_OPTIONS_URI=http://${VIRTUAL_HOST}
+      - DRUSH_OPTIONS_URI=${DRUSH_OPTIONS_URI:-http://${VIRTUAL_HOST}}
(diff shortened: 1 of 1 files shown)


lpeabody:
@jhedstrom is a Drush alias specifies it's own URI, I'm assuming that overrides this option? If so, this addition is rather rad.

jhedstrom:
@lpeabody good question! I'm digging into that and will get back to you. It isn't clear from https://docs.drush.org/en/master/using-drush-configuration/

lpeabody:
Seems like it does override, which would make sense, I was just asking because I'm a paranoid developer.

```
[les.peabody@~/Sites/corvette (feature/core-update)]
-> fin exec 'echo $DRUSH_OPTIONS_URI'
http://domain.docksal

[les.peabody@~/Sites/corvette (feature/core-update)]
-> fin drush @local.com uli
http://com.domain.docksal/user/reset/1/1565310275/MzboziHxPaQYvSQDhKuc29poboV9aJ9bLbBxP-eN7X8/login
```

jhedstrom:
Greg from the drush slack channel also verified this to be the case.

> `DRUSH_OPTIONS_URI` is equivalent to setting the option in a configuration file. Commandline and alias options have a higher priority.

sean-e-dietrich:
@jhedstrom @lpeabody how does this work with multi-sites?

jhedstrom:
@sean-e-dietrich it wouldn't. This is to provide a better default behavior out-of-the-box than `http://default/` for projects that have not yet specified any alias files. Multisite projects would still need to craft alias files I think :)

sean-e-dietrich:
@jhedstrom my suggestion here would be to leave the value empty then and then allow someone to change per project

stacks/services.yml
```diff
@@ -147,6 +147,7 @@ services:
       - XDEBUG_CONFIG=remote_connect_back=0 remote_host=${DOCKSAL_HOST_IP}  # Point xdebug to the host IP
       - PHP_IDE_CONFIG=serverName=${VIRTUAL_HOST}
       - DRUSH_ALLOW_XDEBUG=${XDEBUG_ENABLED:-0}
+      - DRUSH_OPTIONS_URI=http://${VIRTUAL_HOST}
```
sean-e-dietrich:
```suggestion
      - DRUSH_OPTIONS_URI
```

sean-e-dietrich: that way the variable can be customized

jhedstrom:
These changes have been made

sean-e-dietrich:
hi @jhedstrom took a look at this and appears that we can't do variables in variables which is a bummer. I don't want to set to the virtual host for a fear that for multi-sites. I'm wondering if we were to set this as an empty variable and then allow people to configure per project might be a happy medium. thoughts?

lmakarov:
Without any overrides or mappings in `sites.php`, drush will pick the uri based on the site folder name, which is `default` by default.

```
$ drush --version
Drush Launcher Version: 0.6.0
Drush Commandline Tool 9.7.1

docker@cli:/var/www/docroot$ drush st
 Drupal version   : 8.8.0-dev                                     
 Site URI         : http://default
 ...

docker@cli:/var/www/docroot/sites/site1$ drush st
 Drupal version   : 8.8.0-dev                                     
 Site URI         : http://site1

```

`DRUSH_OPTIONS_URI` only works with Drush 9+ and is an equivalent (almost) of `--uri` command line parameter for Drush 8. `--uri` is also supported with Drush 9 and takes precedence over `DRUSH_OPTIONS_URI` as well as aliases (https://github.com/drush-ops/drush/pull/3966)

Eventually, the order of precedence is the following (from most to least important):

- `--uri` on command line
- `uri` in alias
- `DRUSH_OPTIONS_URI`

```
$ DRUSH_OPTIONS_URI="${VIRTUAL_HOST}" drush st
 Drupal version   : 8.8.0-dev                                     
 Site URI         : http://drupal8-contrib.docksal
 ...

$ drush @example.local st
 Drupal version   : 8.8.0-dev                                     
 Site URI         : http://www.example.com
 ...

DRUSH_OPTIONS_URI="${VIRTUAL_HOST}" drush @example.local st
 Drupal version   : 8.8.0-dev                                     
 Site URI         : http://www.example.com

$ DRUSH_OPTIONS_URI="${VIRTUAL_HOST}" drush @example.local st --uri=local.docksal
 Drupal version   : 8.8.0-dev                                     
 Site URI         : http://local.docksal
 ...
```

Sooo... there is absolutely no hard to multisite projects (or anything else for that matter) in setting the `DRUSH_OPTIONS_URI=${VIRTUAL_HOST}` env variable on the cli container.

Issue #751, opened 2018-08-01
Document Startup.sh Script


sean-e-dietrich:
A feature was included as part of the CLI container. Unfortunately docs were never created around it https://github.com/docksal/service-cli/pull/55

PR #1146, opened 2019-10-02
Document the Startup.sh Script

sean-e-dietrich:
Solves #751


docs/content/service/cli/startup.md
```diff
@@ -0,0 +1,15 @@
...
+{{% notice note %}}
+Currently, custom startup scripts are only supported within the `cli` container.
+{{% /notice %}}
+
+## Configuring Startup Scripts
+
+Startup scripts can be used to execute whenever a container is started. Usually this is the need to make sure a service 
+is installed, to making sure a file is compiled, to possibly regenerating a key. This file is located within the projects 
+`.docksal/services/cli` directory. The file name should be `startup.sh` and within there all commands will run at the time 
+of container startup. Addtionally make sure this file has execute permissions is important otherwise it will not run.
```
shelane: Startup scripts can be used to execute actions whenever a container is started. This can be ensuring a service 
is installed, making sure a file is compiled, or regenerating a key. This file is located within the projects 
`.docksal/services/cli` directory and should be named `startup.sh`. All commands in the file will run at the time 
of container startup. Make sure this file has execute permissions or it will not run.

PR #1151, opened 2019-10-03
Added IPv6 localhost termination

### LINKED ISSUE #766: fin hosts add terminate IPv6 Address
When running `fin hosts add` it only adds the IPv4 address but sometimes browsers and routers will attempt to connect to the IPv6 address.

Adding the following to the hosts file should fix this.

```
::1 VIRTUAL_HOST.docksal
```

sean-e-dietrich:
Fixes #766


lmakarov:
@sean-e-dietrich please rebase onto `develop` to have clean tests (tests were fixed in `develop` yesterday).

PR #1175, opened 2019-10-13
adds Drupal 8 BLT version for project create option

### LINKED ISSUE #1011: Add Acquia BLT as a project creation option.
## Description

Add Acquia BLT as an option when creating a new Drupal 8 project. The following Drupal 8 options would exist when creating a new project:

* Drupal 8
* Drupal 8 (Composer Version)
* Drupal 8 (Acquia BLT Version)

This would setup blt.yml configuration to be compatible with Docksal OOTB. I pretty much work exclusively through BLT workflow at this point because of it's robust usage of configuration, creation of deployment artifacts, setup and configuration import strategies, and more.

Related issues https://github.com/docksal/docksal/issues/428 (Make Acquia BLT friends with Docksal). It seems like the near-consensus in that thread was that it would be "weird to initialize Docksal from BLT". In reality, folks have already chosen what their operating environment will be and will initialize a BLT project within that environment on their own. Docksal also knows how to set itself up for any given project better than that project will know how to set itself up for Docksal.

## Blocking Issues

- [ ] When running on MacOS (and I imagine Windows as well) and [following the instructions for creating a new BLT project](https://blt.readthedocs.io/en/latest/creating-new-project/), when run through either `fin rc` or `fin exec` after having created a Docksal project already we're seeing OOM errors when Composer tries to install/update packages. OOM persists even if passing `COMPOSER_MEMORY_LIMIT=-1` to the `create-project` command. `fin rc COMPOSER_MEMORY_LIMIT=-1 composer create-project --no-interaction acquia/blt-project my-project`. On Linux this is a non-issue.

    - Need to attempt with PHP and Composer memory limits completely disabled. Also test by increasing RAM allocated to VM (currently fails with 4GB of RAM allocated).

## Potential Issues

* When running Behat tests and leveraging headless Chrome, BLT expects the `google-chrome` binary to be available on the system running the `behat` command. (It will simply refuse to run `blt tests:behat`).
  * This is something that `docksal/cli` will likely never, nor should it, support. I think the onus is on BLT to support not forcing this. In Docker Compose projects it is trivial to include a headless Chrome service that lives on a host in the project network that you can reference in behat.yml. With the emergence and popularity of Docker Compose based projects (Docksal, Lando, DDev, etc.) BLT should really adapt. They may have changed this requirement in 10.x but I'm not sure.

Additional issues will be added to this list at this proposal is fleshed out.

shelane:
Closes #1011

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -3233,7 +3233,7 @@ project_create ()
 			target_cms="Symfony WebApp"
 			target_host_name_search_string="symfony.docksal"
 			target_repo="$URL_REPO_SYMFONY_WEBAPP"
-		    ;;
+			;;
 		10)
 			target_cms="Grav CMS"
 			target_host_name_search_string=""
--- a/bin/fin
+++ b/bin/fin
@@ -3243,7 +3243,7 @@ project_create ()
 			target_cms="Backdrop CMS"
 			target_host_name_search_string="backdrop.docksal"
 			target_repo="$URL_REPO_BACKDROP"
-		    ;;
+			;;
 		12)
 			target_cms="Hugo"
 			target_host_name_search_string=""
--- a/bin/fin
+++ b/bin/fin
@@ -278,6 +278,7 @@ URL_REPO_UI="https://github.com/docksal/docksal"
 URL_REPO_DRUPAL7="https://github.com/docksal/boilerplate-drupal7.git"
 URL_REPO_DRUPAL8="https://github.com/docksal/boilerplate-drupal8.git"
 URL_REPO_DRUPAL8COMPOSER="https://github.com/docksal/boilerplate-drupal8-composer.git"
+URL_REPO_DRUPAL8BLT="https://github.com/docksal/boilerplate-blt.git"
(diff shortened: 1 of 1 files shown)


shelane:
I also ran a local test of a project create to select this option.

bin/fin
```diff
@@ -3198,61 +3200,66 @@ project_create ()
...
 			target_repo="$URL_REPO_SYMFONY_SKELETON"
 			;;
-		8)
+		9)
 			target_cms="Symfony WebApp"
 			target_host_name_search_string="symfony.docksal"
 			target_repo="$URL_REPO_SYMFONY_WEBAPP"
-		;;
-		9)
+		    ;;
```
```suggestion
			;;
```

bin/fin
```diff
@@ -3198,61 +3200,66 @@ project_create ()
...
 			target_repo="$URL_REPO_GRAV"
 			;;
-		10)
+		11)
 			target_cms="Backdrop CMS"
 			target_host_name_search_string="backdrop.docksal"
 			target_repo="$URL_REPO_BACKDROP"
-		;;
-		11)
+		    ;;
```
```suggestion
			;;
```

PR #1178, opened 2019-10-15
Make health check timeout configurable

### LINKED ISSUE #1157: "Waiting for project stack to become ready..." timeout
Sometimes we have this within CI:

image

How to increase this timeout?

achekulaev:
In certain cases project containers would not start within 60 seconds. Despite usually you better investigate what exactly takes so long, sometimes it takes time for good reasons, that cannot or should not be worked around. 
This allows increasing the timeout.

Closes #1157

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -271,6 +271,9 @@ else
 	DOCKSAL_SSH_AGENT_USE_HOST="${DOCKSAL_SSH_AGENT_USE_HOST:-0}"
 fi
 
+# Healthcheck timeout
+DOCKSAL_HEALTHCHECK_TIMEOUT=${DOCKSAL_HEALTHCHECK_TIMEOUT:60}
+
 #---------------------------- URL references --------------------------------
 GITHUB_API="https://api.github.com"
 URL_REPO="https://raw.githubusercontent.com/docksal/docksal"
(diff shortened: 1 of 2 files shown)


bin/fin
```diff
@@ -271,6 +271,9 @@ else
 	DOCKSAL_SSH_AGENT_USE_HOST="${DOCKSAL_SSH_AGENT_USE_HOST:-0}"
 fi
 
+# Healthcheck timeout
+DOCKSAL_HEALTHCHECK_TIMEOUT=${DOCKSAL_HEALTHCHECK_TIMEOUT:60}
```
```suggestion
DOCKSAL_HEALTHCHECK_TIMEOUT=${DOCKSAL_HEALTHCHECK_TIMEOUT:-60}
```

docs/content/stack/configuration-variables.md
```diff
@@ -128,6 +128,12 @@ This is the domain to use which is tacked on to the end of the projects url.
 
 Allow disabling the DNS resolver configuration (in case there are issues with it). Set to `true` to activate.
 
+### DOCKSAL_HEALTHCHECK_TIMEOUT
+
+`Default: 60`
+
+How many seconds to give project containers before `fin project start` stops waiting for them to reach healthy state. (Should divide by 5)
```
lmakarov:
```suggestion
How many seconds to give project containers to reach `healthy` state before `fin` considers the stack startup as failed. (Should be a multiple of 5)
```

I replaced `fin project start` with `fin`, because we should (and will) have the healthcheck done during restart/reset as well, not just `start`.

PR #1179, opened 2019-10-16
Add --quiet option for fin ssh-key add command

sergey-zabolotny:


achekulaev:
I see what you're trying to do, but this needs a rework. 

1. Pass all arguments to `ssh-add` function instead of parsing out `--quiet` and creating another environment variable. 
2. In `ssh-add` function use `parse_params` to parse params. See examples in code and help on using the function above it.

PR #1193, opened 2019-10-21
adds wordpress settings doc

### LINKED ISSUE #1144: Add Wordpress setup/settings docs
We need a section in docs with instructions for Wordpress setup/settings, similar to how we have a [section for Drupal setting](https://docs.docksal.io/tools/drupal/).

shelane:
I know almost nothing of Wordpress. I have simply looked at the docs for Drupal and looked at the init scripts for the boilerplate Wordpress site. Please update/modify as necessary.

Closes #1144

### DIFF
--- a/docs/content/tools/wordpress.md
+++ b/docs/content/tools/wordpress.md
@@ -1,0 +1,27 @@
+---
+title: "Wordpress settings"
+
+---
+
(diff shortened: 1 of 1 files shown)


achekulaev:
This would need to be checked prior to merging

PR #1199, opened 2019-10-24
Refactor tests

achekulaev:
- Tests run **3x faster**
- Comment out cloud provider tests
- Refactor and make other tests run in parallel
  - Break general.bats into project.bats and commands.bats
  - Remove outdated tests and do some refactoring
- Refactor scripting and make travis.yml simpler


tests/general.bats
```diff
@@ -288,87 +288,6 @@ EOF
 	unset output
 }
 
-@test "fin config" {
```
achekulaev: there are copies of these tests in config.bats

tests/system.bats
```diff
@@ -58,18 +58,6 @@ DOCKSAL_IP=192.168.64.100
 	unset output
 }
 
-@test "DSN: .docksal name resolution via nslookup" {
```
achekulaev: unneeded, since this test was defunct and there is another nslookup test at the end of this file

tests/cli-tools.bats
```diff
@@ -1,42 +0,0 @@
-#!/usr/bin/env bats
```
achekulaev: Relevant tests refactored and moved to system.bats

tests/system.bats
lmakarov: This one tests nslookup on the host. It's beneficial when running tests locally. I'll leave it there.

.travis.yml
```diff
@@ -8,16 +8,19 @@ env:
...
     - DOCKSAL_SSH_AGENT_USE_HOST=0
   matrix:
-    - PROVIDER=none
-    - PROVIDER=acquia
-    - PROVIDER=pantheon
-    - PROVIDER=platform
-    - PROVIDER=drush
-    - PROVIDER=wp
+    #- PROVIDER=quicktest # Uncomment to run
+    - PROVIDER=project
```
lmakarov: Let's rename `PROVIDER` to `GROUP` or `SUITE`

tests/config.bats
```diff
@@ -37,7 +37,7 @@ teardown() {
 	mkdir .docksal
 
 	run fin config env
-	[[ $status == 0 ]] && 
+	[[ $status == 0 ]] &&
```
lmakarov: I believe, ` &&` is not necessary here and can be dropped everywhere.

lmakarov:
Implemented the changes I suggested above. Also, ran tests locally on Mac and fixed a few things. LGTM now.

PR #1214, opened 2019-11-12
Only call ssh_key add in up/restart/reset

lmakarov:
This reverts the change in https://github.com/docksal/docksal/pull/1088/commits/3e6f1649b6c44eea9fe03a40a909f9d0dd18e2ed
ssh_key add is an expensive operation (~1.5s), so we should only call it where absolutely necessary.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -978,10 +978,6 @@ check_docksal_running ()
 		sleep 1
 		system_reset
 	fi
-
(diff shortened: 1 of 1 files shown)


lmakarov:
@GMTA will this cause any issues with the other changes from https://github.com/docksal/docksal/pull/1088?

PR #1219, opened 2019-11-14
Use NFS volumes on Mac by default

### LINKED ISSUE #1025: Use NFS volumes by default with Docker for Mac
This is a followup to #711 and #867.

> docksal misc comparison - google sheets 2018-12-01 01-28-52
> 
> Docker for Mac + NFS delivers the same performance/overhead ratio as with VirtualBox + NFS - about 10% overhead on top of natives file system performance. The only exception is the first ("cold"?) run.
> 
> `osxfs:cached` is about 2x slower than native fs (40-50% slower than NFS).
> `osxfs` (added for completeness) is about 3x slower than native fs.

NFS is the default and battle tested option with VirtualBox.

Since it beats `osxfs:cached` in performance by almost 40%, it makes sense to make it the default option for Docker for Mac installations. In cases where file watchers are more important than FS speed, users will be able to override the default and switch to `osxfs:cached`.

lmakarov:
- Setting `DOCKSAL_VOLUMES=nfs` on Mac (applies to Docker Desktop and VBox)
- Disabled the default NFS mount triggered at VM start for VBox
- Added switching between nfs volume and bind mount in `run_cli`

Closes #1025 

- [x] Tests pass on Docker Desktop
- [x] Tests pass on VirtualBox

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -5288,6 +5288,16 @@ run_cli ()
 		docker volume rm -f ${HOME_VOLUME_NAME} >/dev/null 2>&1
 	fi
 
+	local MOUNT_PWD
+	if [[ "${DOCKSAL_VOLUMES}" == "nfs" ]]; then
+		# Mounts options here should match those in volumes-nfs.yml
+		# Also, see https://docs.docker.com/storage/volumes/#choose-the--v-or---mount-flag
+		# and "Escape values from outer CSV parser" section on that page
+		MOUNT_PWD="type=volume,dst=/var/www,volume-driver=local,volume-opt=type=nfs,volume-opt=device=:$(pwd),\"volume-opt=o=addr=${DOCKSAL_HOST_IP},vers=3,nolock,noacl,nocto,noatime,nodiratime,actimeo=1\""
+	else
+		MOUNT_PWD="type=bind,src=$(pwd),dst=/var/www"
+	fi
+
 	local MOUNT_HOME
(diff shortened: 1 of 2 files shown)


bin/fin
```diff
@@ -3811,9 +3811,11 @@ configure_mounts ()
...
 
 			# Mount the share in VirtualBox
-			if ! is_docker_native; then
-				docker_machine_mount_nfs "${DOCKSAL_NFS_PATH}:${DOCKSAL_NFS_PATH}"
-			fi
+			# TODO: Disabled. To be removed.
+			# NFS mounts are now handled using NFS volumes - same approach as with Docker Desktop
+			#if ! is_docker_native; then
+			#	docker_machine_mount_nfs "${DOCKSAL_NFS_PATH}:${DOCKSAL_NFS_PATH}"
+			#fi
```
lmakarov: I'm not 100% sure whether this should be dropped or kept. The reason for keeping would be to match the state between VBox and Docker Desktop. Docker Desktop has the paths like `/Users` shared inside the VM by default via `osxfs`. I'm wonder if we should just treat VBox the same way and keep the default `/User` NFS mount. @achekulaev thoughts?

lmakarov: Leaving the default NFS share alone for now, since this would be a breaking change for some projects with custom/old stack configs. We also still have docs mentioning the old style usage of `$PROJECT_ROOT:/var/www` in volumes.

PR #1221, opened 2019-11-15
Revamp volumes documentation

achekulaev:


docs/content/core/volumes.md
```diff
@@ -7,35 +7,44 @@ aliases:
 
 ## Quick Overview of the Volumes Options
 
-If option is marked Default, then you do not have to do anything for this option to be used.
-
-If `DOCKSAL_VOLUMES` option is not Default, then you need to enable it manually to use it.
+If `DOCKSAL_VOLUMES` option is not marked as default, then you need to enable it manually to use it.
 All non-automatic options are enabled by placing `DOCKSAL_VOLUMES="<value from volumes column>"` into the respective
```
shelane: Should automatic be changed to default here too?

achekulaev: Thanks, will change now

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
-We can also do more advanced and pretty interesting solutions, like using Unison to synchronize files between the host 
-and the `project_root` volume. 
+To see how your project's Docker volumes are defined with `DOCKSAL_VOLUMES=nfs` see 
+[stacks/volumes-nfs.yml](https://github.com/docksal/docksal/blob/master/stacks/volumes-nfs.yml). 
 
-See [stacks/volumes-unison.yml](https://github.com/docksal/docksal/blob/master/stacks/volumes-unison.yml).
+### unison
 
-Unison volumes make the most sense for Docker for Mac users as an alternative to the (still slow) `osxfs` file sharing.
+This option only works on macOS and Windows. Host files are made accessible to VM like in a bind mount, but after
```
shelane: to the VM

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
+that container does not access them directly via bind to avoid NFS/SMB performance penalty. Container file system is not
```
shelane: that, the container... penalty. The container file system

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
-See [stacks/volumes-unison.yml](https://github.com/docksal/docksal/blob/master/stacks/volumes-unison.yml).
+### unison
 
-Unison volumes make the most sense for Docker for Mac users as an alternative to the (still slow) `osxfs` file sharing.
+This option only works on macOS and Windows. Host files are made accessible to VM like in a bind mount, but after
+that container does not access them directly via bind to avoid NFS/SMB performance penalty. Container file system is not 
+connected to host filesystem at all.   
 
-This is what the file sharing chain looks like with Unison over `osxfs`. 
+Instead `fin` launches a `unison` container with Unison daemon that copies files back and forth between mounted host
```
shelane: Instead,

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
 
-Unison volumes make the most sense for Docker for Mac users as an alternative to the (still slow) `osxfs` file sharing.
+This option only works on macOS and Windows. Host files are made accessible to VM like in a bind mount, but after
+that container does not access them directly via bind to avoid NFS/SMB performance penalty. Container file system is not 
+connected to host filesystem at all.   
 
-This is what the file sharing chain looks like with Unison over `osxfs`. 
+Instead `fin` launches a `unison` container with Unison daemon that copies files back and forth between mounted host 
+files and the container. Transferrring changes from host to the container and back becomes slower, but reading and writing 
+the files within the container becomes way faster.
```
shelane: becomes much faster.

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
+#### Using Unison Volumes
 
-- [Install Docksal](/getting-started/setup/)
 - Add `DOCKSAL_VOLUMES=unison` into `.docksal/docksal.env` of a project
 - `fin project reset`
 - Wait until initial sync finishes.
 
-## None Volumes
+{{% notice note %}}
+Once you set a new `DOCKSAL_VOLUMES` option, you must re-create `cli` container. The easiest way is `fin project reset`,
```
shelane: recreate the cli container

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
+but it will also remove all data from the `db` volume. If you want to retain it then remove only `cli` container, and
```
shelane: retain it, remove only the cli container

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
+{{% /notice %}}
+
+To see how your project's Docker volumes are defined with `DOCKSAL_VOLUMES=unison` see 
+[stacks/volumes-unison.yml](https://github.com/docksal/docksal/blob/master/stacks/volumes-unison.yml).
+
+### none
 
-This method is similar to the Unison method, but without the actual sync happening at all.  
-Nothing is mounted/copied from the host. An empty `project_root` volume is created and mounted inside containers.
+Advanced option. With this option host files are made accessible to VM like in a bind mount, but after that
```
shelane: With this option,
accessible to the VM
but after that,

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
+
+To see how your project's Docker volumes are defined with `DOCKSAL_VOLUMES=unison` see 
+[stacks/volumes-unison.yml](https://github.com/docksal/docksal/blob/master/stacks/volumes-unison.yml).
+
+### none
 
-This method is similar to the Unison method, but without the actual sync happening at all.  
-Nothing is mounted/copied from the host. An empty `project_root` volume is created and mounted inside containers.
+Advanced option. With this option host files are made accessible to VM like in a bind mount, but after that 
+container does not access them directly via bind, nor files are copied over.
```
shelane: the container

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
-This method is similar to the Unison method, but without the actual sync happening at all.  
-Nothing is mounted/copied from the host. An empty `project_root` volume is created and mounted inside containers.
+Advanced option. With this option host files are made accessible to VM like in a bind mount, but after that 
+container does not access them directly via bind, nor files are copied over. 
+You would have to copy files into container and back manually.
 
-Provides **the best** file system performance. Combined with Cloud9, can provide a way of provisioning instant blank 
-development environments with the best performance and consistency for Mac and Windows (Linux has the best performance 
-naturally). The only added cost is having to stick with a web based IDE and terminal.
+Combined with [VSCode IDE](/tools/ide/) this option can provide a way of provisioning instant blank
```
shelane: ), this option

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
   cli:
     volumes:
-      - ${PROJECT_ROOT}:/var/www:rw
+      - /Users/alex/mysite:/var/www:rw
 ```
 
-`${PROJECT_ROOT}` is automatically set to the project's root directory on the host.
-
-The whole mount chain looks like this (drop the last part for Linux hosts).
+It is an unnamed volume with bind driver. However to support various OS in Docksal we cannot just always rely on the
```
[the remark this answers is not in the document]
achekulaev: In Russian it sounds alright xD I think I meant "we just cannot always"

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
     volumes:
-      - ${PROJECT_ROOT}:/var/www:rw
+      - /Users/alex/mysite:/var/www:rw
 ```
 
-`${PROJECT_ROOT}` is automatically set to the project's root directory on the host.
-
-The whole mount chain looks like this (drop the last part for Linux hosts).
+It is an unnamed volume with bind driver. However to support various OS in Docksal we cannot just always rely on the 
+bind driver, and should not use unnamed volumes.
```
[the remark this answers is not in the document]
achekulaev: What about "always comma before and" rule?

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
 
-See [stacks/volumes-bind.yml](https://github.com/docksal/docksal/blob/master/stacks/volumes-bind.yml).
+`DOCKSAL_VOLUMES` value changes the driver option for project volumes mentioned above, and also affects some additional `fin` behavior. 
 
-Defining volumes this way makes it much easier to override volume settings in one place (`volumes` section) vs multiple 
-places in the yaml file. We can now swap bind mounting with something else. See below.
+### bind 
 
-### osxfs:cached Mode with Docker for Mac
+With this option containers access files using **Docker bind** driver option, which basically means direct access.
```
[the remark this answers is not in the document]
achekulaev: sorry?

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
-Defining volumes this way makes it much easier to override volume settings in one place (`volumes` section) vs multiple 
-places in the yaml file. We can now swap bind mounting with something else. See below.
+### bind 
 
-### osxfs:cached Mode with Docker for Mac
+With this option containers access files using **Docker bind** driver option, which basically means direct access. 
 
-As explained above for bind volumes option the files are actually mapped to the VM that Docker runs in
-via `osxfs:cached`. Docksal automatically enables the `osxfs:cached` mode on Docker for Mac.
+While Docker can access files directly on Linux, on macOS and Windows it works inside VM (VirtualBox or xhyve with Docker Desktop),
```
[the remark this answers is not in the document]
achekulaev: it's not a specific VM in this case

docs/content/core/volumes.md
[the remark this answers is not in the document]
shelane: Comma after option: With this option, containers

docs/content/core/volumes.md
achekulaev: what's wrong with "way faster"?

docs/content/core/volumes.md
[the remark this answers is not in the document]
shelane: It is specific as far as it is the VM operating. “A” VM (for something generic) doesn’t work because you are specifying an operation. It’s just a quirky American English thing that requires a definite article.

docs/content/core/volumes.md
shelane: There is nothing “wrong” except that it’s more slang. “Much” would just be the preferred.

docs/content/core/volumes.md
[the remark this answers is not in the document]
achekulaev: There is another word option that has comma after it already, so comment was confising

docs/content/core/volumes.md
```diff
@@ -5,228 +5,176 @@ aliases:
...
+macOS Host ==============> Container  
+```
 
-## Unison Volumes
+NFS generally works faster than `osxfs`, so this option is default on macOS even for Docker Desktop setup. The downside
+is that NFS does not support fsnotify events.
 
-We can also do more advanced and pretty interesting solutions, like using Unison to synchronize files between the host 
-and the `project_root` volume. 
+To see how your project's Docker volumes are defined with `DOCKSAL_VOLUMES=nfs` see
```
shelane: =nfs, see

docs/content/core/volumes.md
```diff
@@ -27,17 +27,17 @@ If you are familiar with Docker Compose, then you must have attached files or di
       - /Users/alex/mysite:/var/www:rw
 ```
 
-It is an unnamed volume with bind driver. However, to support various OS in Docksal, we cannot always rely on the 
-bind driver and should not use unnamed volumes. 
+That is a called a bind mount - mounting a files/folder on the host directly into a container.
```
achekulaev: Let's return to what it was here. First the change didn't bring any new info. Second I deliberately avoided calling attaching volumes "mounting" to avoid confusing people who are new that everything is mounting. Mounting is used to describe a process of accessing host files from VM.

docs/content/core/volumes.md
```diff
@@ -48,71 +48,69 @@ If you need to move the location of a project on the computer, remove the projec
...
 
 ## DOCKSAL_VOLUMES
 
-`DOCKSAL_VOLUMES` value changes the driver option for project volumes mentioned above and also affects some additional `fin` behavior. 
+`DOCKSAL_VOLUMES` value changes the driver option for project volumes mentioned above and also affects `fin` behavior. 
 
 ### bind 
 
-With this option, containers access files using **Docker bind** driver option, which basically means direct access. 
+With this option, containers access files via a [bind mount](https://docs.docker.com/storage/bind-mounts/), which basically means direct access.
```
achekulaev: Again let's not call it mount.

docs/content/core/volumes.md
```diff
@@ -5,228 +5,179 @@ aliases:
...
+to a service. Mixing `cached` and non-`cached` mounts for the same volume in your project stack will lead to issues 
+and errors with Docker Desktop on macOS. See [docksal/docksal#678](https://github.com/docksal/docksal/issues/678) for more details.
+{{% /notice %}}
 
-In the example above, `project_root` and `docksal_ssh_agent` are "named volumes." The first one is a project level one,
-while the second one is a global volume and is used by all projects.
+## DOCKSAL_VOLUMES
 
-See [stacks/volumes-bind.yml](https://github.com/docksal/docksal/blob/master/stacks/volumes-bind.yml).
+`DOCKSAL_VOLUMES` value changes the driver option for project volumes mentioned above and also affects `fin` behavior.
```
achekulaev:
```suggestion
`DOCKSAL_VOLUMES` value changes the mount type for project volumes mentioned above and also affects `fin` behavior. 
```

docs/content/core/volumes.md
```diff
@@ -5,228 +5,179 @@ aliases:
...
-Defining volumes this way makes it much easier to override volume settings in one place (`volumes` section) vs multiple 
-places in the yaml file. We can now swap bind mounting with something else. See below.
+### bind 
 
-### osxfs:cached Mode with Docker for Mac
+With this option, containers access files via a [bind mount](https://docs.docker.com/storage/bind-mounts/), which basically means direct access. 
 
-As explained above for bind volumes option the files are actually mapped to the VM that Docker runs in
-via `osxfs:cached`. Docksal automatically enables the `osxfs:cached` mode on Docker for Mac.
+While Docker can access files directly on Linux, on macOS and Windows it works inside a VM (VirtualBox, xhyve/Hyper-V
```
achekulaev:
```suggestion
While Docker can access files directly on Linux, on macOS and Windows it works inside the VM (VirtualBox, xhyve/Hyper-V 
```

As pre @shelane it is "the" here.

docs/content/core/volumes.md
```diff
@@ -5,228 +5,179 @@ aliases:
...
 
 ```
-container:/var/www ==bind mount==> project_root ==> Linux:project_root ==NFS==> macOS:PROJECT_ROOT
+               SMB mount                     bind mount
+Windows Host =============> Docker Desktop ==============> Container   
 ```
 
-As you can see, containers mount NFS via the host machine and not directly. This setup method only makes sense on macOS 
-with Docker for Mac, for testing and performance comparison purposes.  
+To see how your project's Docker volumes are defined with when `DOCKSAL_VOLUMES=bind`, see
```
achekulaev: Remove with or when

docs/content/core/volumes.md
```diff
@@ -5,228 +5,179 @@ aliases:
...
 ---
 
-## Quick Overview of the Volumes Options
+## Quick Overview
 
-If option is marked Automatic, then you do not have to do anything for this option to be used.
+| OS      | VM              | DOCKSAL_VOLUMES              | Comments  |
+|---------|-----------------|------------------------------|-----------|
+| Linux   | -               | [bind](#bind) (**default**)  | Direct host files access, maximum filesystem speed. |
+| macOS   | VirtualBox      | [nfs](#nfs)  (**default**)   | **Pros:** fast, ~20% slower than native filesystem. <br> **Cons:** does not support filesystem events (fsnotify). |
```
achekulaev:
```suggestion
| macOS   | VirtualBox      | [nfs](#nfs)  (**default**)   | **Pros:** fast, 80% of speed of the native filesystem. <br> **Cons:** does not support filesystem events (fsnotify). |
```

Мне всегда как-то странно читалась фраза "быстрый, на 20% медленнее". Предлагаю перефразировать.

docs/content/core/volumes.md
```diff
@@ -5,228 +5,179 @@ aliases:
...
+| macOS   | Docker Desktop  | [nfs](#nfs)  (**default**)   | **Pros:** fast, ~20% slower than native filesystem. <br> **Cons:** does not support filesystem events (fsnotify). |
+| macOS   | Docker Desktop  | [bind](#bind)                | **Pros:** supports filesystem events. <br> **Cons:** pretty slow, 40% slower than native filesystem. |
+| Windows | ANY             | [bind](#bind) (**default**)  | **Pros:** ~50% slower than native filesystem. <br> **Cons:** does not support filesystem events (fsnotify). |
```
achekulaev:
```suggestion
| Windows | ANY             | [bind](#bind) (**default**)  | **Pros:** ~50% slower than native filesystem. <br> **Cons:** does not support filesystem events (fsnotify). |
```

Тут было 20%, откуда 50?

docs/content/core/volumes.md
```diff
@@ -5,228 +5,179 @@ aliases:
...
   cli:
     volumes:
-      - ${PROJECT_ROOT}:/var/www:rw
+      - /Users/alex/mysite:/var/www:rw
 ```
 
-`${PROJECT_ROOT}` is automatically set to the project's root directory on the host.
-
-The whole mount chain looks like this (drop the last part for Linux hosts).
+That is a called a bind mount - mounting a files/folder on the host directly into a container.
```
achekulaev:
```suggestion
That is a called a bind mount - mounting a file/folder directly into a container. 
```

Let us avoid confusing user about what "host" is in terms of this documentation. You and I understand there are several "hosts", but to make the whole read clear to user in this doc lets call "host" a computer that runs Docker or VM. In this case it is not correct that bind mount example above would mount files from host, as on macOS, Windows it is not from host.

docs/content/core/volumes.md
```diff
@@ -5,228 +5,179 @@ aliases:
...
+to a service. Mixing `cached` and non-`cached` mounts for the same volume in your project stack will lead to issues 
+and errors with Docker Desktop on macOS. See [docksal/docksal#678](https://github.com/docksal/docksal/issues/678) for more details.
+{{% /notice %}}
 
-In the example above, `project_root` and `docksal_ssh_agent` are "named volumes." The first one is a project level one,
-while the second one is a global volume and is used by all projects.
+## DOCKSAL_VOLUMES
 
-See [stacks/volumes-bind.yml](https://github.com/docksal/docksal/blob/master/stacks/volumes-bind.yml).
+`DOCKSAL_VOLUMES` value changes the mount type for project volumes mentioned above and also affects `fin` behavior.
```
achekulaev:
```suggestion
`DOCKSAL_VOLUMES` value changes the mount type of the project volumes mentioned above and also affects `fin` behavior. 
```

docs/content/core/volumes.md
lmakarov: Revamped the comparison table

docs/content/core/volumes.md
lmakarov: Revamped the comparison table

PR #1224, opened 2019-11-15
Added variable to configure Composer variables

sean-e-dietrich:
Added ability for a person to configure the Composer Environment variables.


lmakarov:
I think we should just set `COMPOSER_MEMORY_LIMIT=-1` in `docksal/cli` to remove the memory limit for Composer. This way, it's one less memory limit setting that needs to be adjusted, the other two being PHP's `memory_limit` and then the VM memory limit.

The other variables (`COMPOSER_ALLOW_XDEBUG` and `COMPOSER_DISABLE_XDEBUG_WARN`) make sense to me.

lmakarov:
https://github.com/docksal/service-cli/issues/155

sean-e-dietrich:
Sounds good

PR #1230, opened 2019-11-20
NFS update safeguard for Mac users

lmakarov:
- Set `DOCKSAL_VOLUMES=bind` for existing Mac users (VirtualBox and Docker Desktop)
  - This ensures we do not break existing project stacks by switching to nfs volumes as the default
- Show release notes URL after update
  - Open release notes URL on Mac after update

TODOs:
- [x] Manual testing on Mac

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -278,6 +278,7 @@ DOCKSAL_HEALTHCHECK_TIMEOUT=${DOCKSAL_HEALTHCHECK_TIMEOUT:-60}
 GITHUB_API="https://api.github.com"
 URL_REPO="https://raw.githubusercontent.com/docksal/docksal"
 URL_REPO_UI="https://github.com/docksal/docksal"
+URL_RELEASE_NOTES="https://docksal.io/updates"
 URL_REPO_DRUPAL7="https://github.com/docksal/boilerplate-drupal7.git"
 URL_REPO_DRUPAL8="https://github.com/docksal/boilerplate-drupal8.git"
 URL_REPO_DRUPAL8COMPOSER="https://github.com/docksal/boilerplate-drupal8-composer.git"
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -4943,9 +4952,27 @@ update ()
...
 	rm -f "$CONFIG_DOWNLOADS_DIR/*" &>/dev/null
 
+	# Set DOCKSAL_VOLUMES globally for existing and new users on Macs (VirtualBox and Docker Desktop)
+	# Existing users get "bind" to match the previous default behavior.
+	# This ensures we do not break existing project stacks by switching to nfs volumes as the default (1.13.0 release)
+	# New users get "nfs" - the new default
+	if is_update && is_mac && [[ $(config_get --global DOCKSAL_VOLUMES) == "" ]]; then
+		config_set --global DOCKSAL_VOLUMES=bind
+	else
+		config_set --global DOCKSAL_VOLUMES=nfs
```
achekulaev: Это надо грохнуть. Если любое из условий не соблюдено то делать просто ничего не надо

bin/fin
```diff
@@ -4864,6 +4865,14 @@ is_update ()
 # Update Docksal
 update ()
 {
+	if is_mac; then
+		echo-warning "Please take a few minutes to review the release notes:" \
+			"${URL_RELEASE_NOTES}"
+		open "${URL_RELEASE_NOTES}"
```
achekulaev: Зачем два раза-то?

lmakarov: See updated code. The message is for everyone. Macs also get a page open in browser.

lmakarov: Actually... this was a copy/paste mistake. I will update.

bin/fin
lmakarov: Revised

PR #1234, opened 2019-11-25
Alpine fixes

sergey-zabolotny:


bin/fin
```diff
@@ -4437,7 +4437,7 @@ install_tools_alpine ()
 		echo-green "Installing Docker Compose..."
```
```suggestion
		echo-green "Installing Docker Compose via pip (this may take a while)..."
```

achekulaev:
@lmakarov this will require testing it does not break ubuntu

lmakarov:
@achekulaev the two changes in this PR only affect Alpine. Also, already tested this on Ubuntu 18.04.1 LTS :)

PR #1247, opened 2019-12-17
Improve performance and functionality of `fin bash`

GMTA:
When running `fin bash`, a special value was passed to `_exec()` which sets the
command to run to an interactive shell. This caused the shell and login scripts
to be started twice, which causes an unnecessary delay before the shell is
available for the user.

This commit changes `_exec()` in the following ways:

* Instead of executing `cd {path}` to switch the working directory, use the
  `--workdir` option of `docker exec` instead.
* Consider `__SHELL_INTERACTIVE__` a special case where we only run an
  interactive shell without passing a command to run. This cuts time in half
  for a `fin bash` invocation.
* Working directory is now also changed for the `web` container (when running
  `fin bash web`).

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -5114,91 +5114,86 @@ _exec ()
 	fi
 
 	# Allow overriding container, where to run
-	local container_name
+	local container_name="cli"
 	if [[ "$1" =~ "--in" ]]; then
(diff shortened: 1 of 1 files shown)


achekulaev:
@GMTA this is an awesome work!

PR #1255, opened 2019-12-23
Xhprof Feature

sean-e-dietrich:
Added feature for Xhprof to be enabled. 

This is dependent on the following PR https://github.com/docksal/service-cli/pull/166


docs/content/stack/configuration-variables.md
```diff
@@ -41,7 +41,7 @@ The default amount of disk space allocated to Docksal's VirtualBox machine, in m
 
 `Default: 0`
 
-Applicable to macOS or Windows.
```
lmakarov: Looks like your IDE is configured to remove extra spaces. Spaces may be important. Like this  double space that breaks the line. Please revert the irrelevant changes that were pushed along in this PR.

stacks/overrides-xhprof.yml
```diff
@@ -0,0 +1,34 @@
...
+  cli:
+    environment:
+      - XHPROF_ENABLED=1
+      - XHPROF_OUTPUT_DIR=${XHPROF_OUTPUT_DIR:-/tmp/xhprof}
+
+  # XHProf Service to review
+  xhprof:
+    image: ${XHPROF_IMAGE:-wodby/xhprof}
+    hostname: xhprof
+    expose:
```
lmakarov: Expose is not necessary here, since the image already does that - https://github.com/wodby/xhprof/blob/master/Dockerfile#L36

lmakarov:
@sean-e-dietrich I believe the `xhprof_root` volume needs to be mounted in `cli` to share the xhprof logs with the added `xhprof` viewer service.

sean-e-dietrich:
@lmakarov wanted to bring this back up for review

PR #1263, opened 2020-01-13
Provide database argument when running `fin db cli`

GMTA:
When running `fin db cli`, the default MySQL database is not selected. This only happens when extra arguments are provided.

This PR changes this behavior by always starting `mysql` with the database argument set to whatever `$db` or `$MYSQL_DATABASE` contains.

Note that this is MySQL/MariaDB-specific, but currently `fin db` only deals with those types of databases.


achekulaev:
Looks logical. Thanks

PR #1273, opened 2020-01-30
Correct the creation of the named volumes

ausy-kimkennof:
Use underscores (_) instead of dash (-)


achekulaev:
@ausy-kimkennof some explanations on the reasons of the change are required here.

ausy-kimkennof:
It should be obvious to the reader when you take into account the following lines to update the docksal.yml.

Those following lines use underscores "_".

Issue #498, opened 2018-04-05
New local dev domain + valid HTTPS cert


lmakarov:
Add support for `*.x.docksal.io` with a valid cert from LetsEncrypt.

![image](https://user-images.githubusercontent.com/1205005/38397418-5d90b8d0-38f2-11e8-8d47-ec11c1702402.png)

This will be optional and in addition to the default `*.docksal` domain.

This requires changes to `docksal-vhost-proxy` and `docksal-dns` services.

Related: https://github.com/docksal/docksal/issues/1215

frederickjh:
If it is also possible can we structure the code so that if DOCKSAL_DNS_DOMAIN is set it is also possible to create a LetsEncrypt certificate. My use case for this is, I use Docksal on a development server that uses a domain name that is publicly accessible to allow clients to preview sites.

lmakarov:
@frederickjh are you asking for a fully automated integration with LetsEncrypt? E.g. just set `DOCKSAL_DNS_DOMAIN` and have Docksal automatically request certs from LetEncrypt and configure everything?

Something like that would be a killer feature, indeed. However, it will also require a substantial amount of work. For wildcard domains/certs that may not even be possible to fully automate, since LetsEncrypt requires DNS level verification to issue those.

What will be possible - is do the cert request manually and then have `docksal-vhost-proxy` use it. LetsEncrypt certs are valid for 90 days. After that you'll have to go through the same process again.

frederickjh:
The manual renewal is OK for me. 
Cpanel  and Webmin have ways that you can have LetsEncrypt certificates renewed on a timed basis say every 60 days. They also email when they renew or fail to do so.

lmakarov:
Apparently, the idea of shipping a cert with the app, which points to a local/internal IP is not new.
Also apparently, this is considered bad and the certificate authority is supposed to revoke such certs.

https://letsencrypt.org/docs/certificates-for-localhost/

[...]

PR #1274, opened 2020-02-10
WIP use mkcert for locally-trusted development certificates #498

fjmk:
In #498, [comment](https://github.com/docksal/docksal/issues/498#issuecomment-514158964) is mentioned that we could use [mkcert](https://github.com/FiloSottile/mkcert) for locally-trusted development certificates.

In this pull request I install mkcert, run 'mkcert -install' to create the locally-trusted CAROOT cerificate and create a site cerificate at 'fin project start'.
This certificate is trusted for $VIRTUAL_HOST as well as for *.$VIRTUAL_HOST.

TODOS:
Who can help me to install mkcert on windows and wsl?

- install_tools_wsl  
  Can I use $URL_MKCERT_NIX and MKCERT_BIN="/usr/local/bin/mkcert"
- install_tools_windows  
  Use $URL_MKCERT_WIN and what should MKCERT_BIN_WIN be?
- small change in [service-vhost-proxy](https://github.com/docksal/service-vhost-proxy) to redirect port 80 automatically to 443.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -142,6 +142,7 @@ REQUIREMENTS_VBOX='5.2.34'  # Remember to update the download URLs below
 REQUIREMENTS_WINPTY='0.4.3'
 REQUIREMENTS_WINPTY_CYGWIN='2.8.0'
 REQUIREMENTS_DOCKER_DESKTOP='2.0.0.3'
+REQUIREMENTS_MKCERT='1.4.1'
 
 # VirtualBox download URLs
 # Remember to update REQUIREMENTS_VBOX above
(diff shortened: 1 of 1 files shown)


achekulaev:
It seems that it could play better as an addon maybe? Like one where user would explicitly run `fin mkcert` or such, and only then have certificate generated for the project.

lmakarov:
As much as I would like to have this streamlined for our users, automatically installing `mkcert` and creating a locally trusted CA may raise (valid) security concerns. This should be an optional step and require a consent from the user. 

I'm not yet sure whether this should go in core or be an addon. As an addon, it would have to be a global addon (not a project level addon), which I believe we intentionally do not support today. This may be the case to reconsider support for global addons.

achekulaev:
We totally support global addons. Nothing wrong with them.

lmakarov:
> We totally support global addons. Nothing wrong with them.

@achekulaev We support [global custom commands](https://docs.docksal.io/fin/custom-commands/#global-custom-commands), but not global addons. Addons are project specific and cannot be installed globally using `fin` today.

achekulaev:
@lmakarov global addons will work if placed where they should be. Installation is a separate question.

fjmk:
I understand your reservations.

I submitted an [addon mkcert](https://github.com/docksal/addons/pull/54) pull request with a lot of questions....

lmakarov:
The addon is here: https://github.com/docksal/addons/tree/master/mkcert
Issue to include it in docs.docksal.io: https://github.com/docksal/docksal/issues/1370

PR #1283, opened 2020-02-23
Update for Elasticsearch 7 compatibility

phpsa:
I needed to use Elasticsearch 7 for one of my projects and could not get it to start by default 
Found that I needed to use the following in my override Yaml file:
```yaml
elasticsearch:
    extends:
      file: ${HOME}/.docksal/stacks/services.yml
      service: elasticsearch
    image: elasticsearch:7.1.1
    environment:
      - bootstrap.memory_lock=true
      - discovery.type=single-node
      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
```
Whereas it should be easier to use 
```bash
ELASTICSEARCH_IMAGE=elasticsearch:7.1.1
```
in my env file

Single node addon was also needed in order to allow the node to start

### DIFF
--- a/stacks/services.yml
+++ b/stacks/services.yml
@@ -245,7 +245,7 @@ services:
   # Elastic Search
   elasticsearch:
     hostname: elasticsearch
-    image: elasticsearch:6.5.3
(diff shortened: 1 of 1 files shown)


lmakarov:
Thanks!

PR #1289, opened 2020-02-27
Fix environment variables for `fin exec` running scripts

GMTA:
I discovered some issues with `fin exec` when executing scripts. There is a different flow when the first argument to `fin exec` is an existing file which has two issues:

1. The `$1` argument is duplicated to `$script_to_run` but no `shift` is executed, resulting in the `$1` argument being duplicated as part of `$@`
2. There are environment variables passed to `_exec` which result in garbled variables since the entire string is pulled through `printf`. I see no reason to define these variables again since they are also configured as environment variables in the Docker container itself. 

In the end, this makes things a bit simpler and fixes the usage of the `$PROJECT_ROOT` variable when executing scripts with `fin exec`.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -8034,7 +8034,8 @@ case "$1" in
 			else
 				script_to_run="$1"
 			fi
-			_exec "PROJECT_ROOT=/var/www DOCROOT=$DOCROOT VIRTUAL_HOST=$VIRTUAL_HOST /bin/bash -c $script_to_run" "$@"
+			shift
+			_exec "$script_to_run" "$@"
 		else
 			_exec "$@"
 		fi


achekulaev:
I don't think it's correct. 
I haven't even tested it by running `fin exec 'echo $DOCROOT'` which might fail after this change, but double `shift` does not seem correct at all.

GMTA:
@achekulaev 

> I don't think it's correct.
> I haven't even tested it by running `fin exec 'echo $DOCROOT'` which will probably fail after this change

Works perfectly for me?

```
% fin exec 'echo $PROJECT_ROOT'
/var/www
% fin exec 'echo $DOCROOT'
public
% fin exec 'echo $VIRTUAL_HOST'
example-app.docksal
```

> but double `shift` does not seem correct at all.

The first `shift` removes the exec/run argument. The second `shift` removes the filename (stored in `$script_to_run`) so `$@` does not expand with the filename as the first argument.

But I'll gladly try to fix anything that theoretically doesn't work ;-)

achekulaev:
Ok I'll try to merge it and let's see how it goes. It only affect the case with file anyways.

Issue #1288, opened 2020-02-26
CLI Container startup fails sometimes on lua script project start


ey-:
**Description**

_Sometimes_ the project startup fails when the project is on a paused state and you hit the url via a browser. All containers are started without an issue, except the cli container.
Following can be found in the logs for the vhost-proxy when this happens:

```
2020-02-26 22:56:52 [proxyctl] [lookup]
2020-02-26 22:56:52 [proxyctl] [lookup] Found a project stack 'my-project'...
2020-02-26 22:56:52 [proxyctl] [start]
2020-02-26 22:56:52 [proxyctl] [start] Starting project stack for 'my-project'...
2020-02-26 22:56:52 [proxyctl] [start] Connected proxy to network 'my-project_default'.
2020-02-26 22:56:53 [proxyctl] [start] Stopping docker-gen...
2020-02-26 22:56:53,354 INFO waiting for docker-gen to stop
2020/02/26 22:56:53 Received signal: terminated
2020/02/26 22:56:53 Received signal: terminated
2020-02-26 22:56:53,356 INFO stopped: docker-gen (exit status 0)
docker-gen: stopped
2020-02-26 22:56:53 [proxyctl] [start] Starting containers for my-project...
2020-02-26 22:56:53 [proxyctl] [start] Pass #1...
5bb4250c1b4e
Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "process_linux.go:430: container init caused \"rootfs_linux.go:58: mounting \\\"/tmp/ssh-XYZQWER/agent.4555\\\" to rootfs \\\"/var/lib/docker/overlay2/SOMEHasH/merged\\\" at \\\"/var/lib/docker/overlay2/SOMEHasH/merged/tmp/ssh-XYZQWER/agent.4555\\\" caused \\\"not a directory\\\"\"": unknown: Are you trying to mount a directory onto a file (or vice-versa)? Check if the specified host path exists and is the expected type
4854468cc19d
14c75c4daf1b
9f2709ff8492
73628b2e1bb8
Error: failed to start containers: 6c69526e9ba9
crond: USER root pid 13852 cmd /usr/local/bin/proxyctl cron
2020-02-26 22:57:00 [proxyctl] [start] Pass #2...
2020-02-26 22:57:00 [proxyctl] [cron]
2020-02-26 22:57:01 [proxyctl] [stop]
5bb4250c1b4e
[...]

PR #1291, opened 2020-03-04
Fix issue with starting cli from vhost-proxy and missing ssh agent socket

sergey-zabolotny:
Fixes: https://github.com/docksal/docksal/issues/1288


lmakarov:
I think we can use `~/.docksal/tmp/${COMPOSE_PROJECT_NAME_SAFE}` for shared directories. This way, we keep Docksal stuff in one place - under `~/.docksal`

PR #1297, opened 2020-03-16
Make cli_home volume name dynamic (based on cli image name/version)

sergey-zabolotny:
Fixes: https://github.com/docksal/docksal/issues/787

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -1302,6 +1302,12 @@ _remove_containers ()
 
 		# Taking the whole docker-compose project down (this removes containers, volumes, and networks)
 		docker-compose down --volumes --remove-orphans
+
+		# Get project orphan volumes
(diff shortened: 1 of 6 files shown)


bin/fin
```diff
@@ -6176,6 +6182,10 @@ load_configuration ()
 	# https://docs.docker.com/compose/reference/envvars/#compose_convert_windows_paths
 	export COMPOSE_CONVERT_WINDOWS_PATHS=1
 
+	# get cli image version
+	CLI_IMAGE_VER=$(docker-compose --verbose config --hash="cli" 2>&1 | grep 'docker inspect_image <-' | cut -d"'" -f2 | sed 's#/#-#g;s#:#-#g')
```
lmakarov, line 6186: This adds almost 2s to `load_configuration()`, which is used in many places in fin.
So this would slow everything down substantially. We need to figure out a different approach.

PR #1302, opened 2020-03-20
Performance: cache `DOCKER_RUNNING` inside `is_docker_running`

GMTA:
I noticed that running a simple `fin` was hanging sometimes when Docker Desktop was still booting. I wasn't sure why this would happen since `fin` does not have to interact with Docker at that point, so I investigated.

The variable `DOCKER_RUNNING` is seeded with Docker's running state for _every_ invocation of the `fin` command. I decided to apply caching inside the `is_docker_running` method and clear `DOCKER_RUNNING` whenever the value needs to be updated. This way, it's only checked if Docker is running if the logic requires it and `fin` no longer hangs when Docker is still booting.

On my machine, this reduces the wall time for `fin` from 310ms to 250ms on average.

Also looking for feedback. I think `is_docker_running` could benefit from a cleaner implementation than the one I suggest in this MR.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -263,6 +263,9 @@ PATH="$CONFIG_BIN_DIR:$PATH"
 # Set global variable in case native Docker app is used/not-used
 DOCKER_NATIVE="${DOCKER_NATIVE:-0}"
 
+# Whether Docker is running - cached value
+DOCKER_RUNNING=""
+
 # SSH agent settings
 if is_ci; then
 	# Prefer the host's agent in CI environments
@@ -791,16 +794,22 @@ check_winpty_found ()
 # Returns 255 - docker-machine is down
 is_docker_running ()
 {
+	[ -n "${DOCKER_RUNNING}" ] && return $DOCKER_RUNNING
+
 	# Running "docker version" is sufficient to check if docker is running in any host environment setup.
 	# This operation is instant even if docker is not running (assuming a socket is used).
 	# However, if DOCKER_HOST is an HTTP/TLS endpoint and it is down, then this check will take over a minute to fail.
 	# To mitigate this case, we first check for is_docker_machine_running when in docker-machine mode.
+	local result=1
 	if ! is_linux && ! is_docker_native; then
-		# Return a distinct code if the machine is down.
-		# This is checked down the road.
-		is_docker_machine_running || return 255
+		# Return a distinct code if the machine is down. This is checked down the road.
+		is_docker_machine_running && result=0 || result=255
+	elif docker version &>/dev/null; then
+		result=0
 	fi
-	docker version &>/dev/null || return 1
+
+	DOCKER_RUNNING=$result
+	return $result
 }
 
 is_docksal_running ()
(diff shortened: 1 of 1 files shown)


achekulaev:
As a known phrase goes, caching (and clearing the caches properly in all corner cases) is one of the most problematic tasks in programming, along with naming things.

I foresee many issues with caching this value (that is indeed calculated every time in all those unforeseen scenarios, that I can't even think of right now. (And I know for sure there will be some, because I implemented caching of Vbox VM running state before, and met those corner cases for that one)

The gain is to prevent `fin` from hanging when Docker is starting (very rare case), and saving 100ms.

The gain doesn't seem worth the future headache of debugging those potential corner cases to me.

GMTA:
@achekulaev I agree with you, however, the caching was already in place. I just centralized it to the `is_docker_running` method and by doing so, improved performance and prevented hangs in some cases.

Would you rather remove the caching altogether?

achekulaev:
Okay sorry I have hard time reading your change request maybe. I understand that you tried to follow "exit code" approach, but really it's very hard to read. I used it in the past but for past year or so all such things are done in more human way with 0 for false and 1 for true. If you can change to those maybe I would better understand what has changed and where the improvement comes from. 

Because improvements should also show if there is some inefficiency right now. If we check for running docker twice. But I don't see where it happens.

GMTA:
Hi @achekulaev,

1. The 0/1/255 exit code approach is something that the method already implemented so it was not added by me. I made it backwards compatible since I do not know the intricacies of Docker machine management in Docksal.
2. The performance improvement achieved here is that previously, it was always checked whether Docker was running. After this change, it is only checked with an invocation to `is_docker_running`. I agree with you that it is hard to find a case where `is_docker_running` and/or `DOCKER_RUNNING` are used twice in one `fin` run. Maybe we would be better of dropping the caching altogether?

achekulaev:
_> I agree with you that it is hard to find a case where is_docker_running and/or DOCKER_RUNNING are used twice in one fin run_

Then I am struggling to understand the reasoning for the PR. It will not help in case when DD is still starting, which seems to be where you started the explanations. 
Which case are you trying to improve then?

GMTA:
@achekulaev 

> Which usecase are you trying to improve then?

There are two outlined (in disguise) in the OP:

> This way, it's only checked if Docker is running if the logic requires it and fin no longer hangs when Docker is still booting.

This describes two improvements:

1. The use cases where you execute `fin` but do not touch the Docker daemon (think `fin`, `fin config`, etc.) benefit since there no longer is any communication with the daemon (performance improvement)
2. When the Docker daemon is booting up, `fin` commands that do not interact with the daemon can arguably be considered broken since they hang until a timeout or successful boot takes place. This changes it in a way these commands will continue working, even is the daemon is not running yet (bugfix)

The caching was refactored since it was already in there. If you think it should be removed, let me know.

achekulaev:
I see thanks. 

But I think you forgot important part. DOCKER_RUNNING is a variable that can be (and is) used in commands/addons. E.g. https://github.com/docksal/addons/blob/master/mailhog/mailhog.pre-install#L4

So twi important takeaways of this: the format of the variable should not be changed and the variable should continue to be calculated for those cases.

GMTA:
@achekulaev Ah, unfortunately that means that we cannot optimize it away. I'll close this MR.

PR #1305, opened 2020-04-01
ensure share works with env variables set

sloped:
Previously if you set docksal.env vars for `ngrok` sharing failed as we did not set the host and port.
This fixes `fin share` those arguments are set even if `docksal.env` variables are set

This is a smaller change that fixes  #1047. That seems to be part of a larger rewrite of ngrok and I am hoping this can be merged and released as a minor release to make `fin share` useable for those who want to use ngrok authentication or other options.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -5794,8 +5794,10 @@ ngrok_share ()
 
 	# Add -host-header by default if no arguments are included.
 	if [[ "${ARGS}" == "ngrok http" ]]; then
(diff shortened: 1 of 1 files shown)


achekulaev:
@sean-e-dietrich how does this PR look to you?

sean-e-dietrich:
The problem with this is that you if you specify a config file you can't specify the container and port it maps to.

lmakarov:
Closing per the explanation in the previous comment.

PR #1306, opened 2020-04-03
Fix database selection when running `db cli` without arguments

GMTA:
Previously, when running `fin db cli` the database parameter was not passed in
causing a MySQL CLI interface to open up without a database selected. This
fixes it by streamlining the invocation of `mysql`.

Changes include:
- Centralize a local `ptyCommand` variable to get rid of a separate `mysql` invocation
- Use optional expansion for the `-e` argument to get rid of another `mysql` invocation
- Use shell argument escaping for username, password and database name

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -5420,17 +5420,16 @@ _mysql ()
 	# MYSQL_* variable expansion happens here
 	local __dump_password="${dbpassword:-$MYSQL_ROOT_PASSWORD}"
 	local __database="${db:-$MYSQL_DATABASE}"
-
-	if [[ "${ARGV[*]}" != "" ]]; then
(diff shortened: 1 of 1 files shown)


achekulaev:
First "ptyCommand" does not match codestyle. Second, I don't see how the case when there is no args is always processed as tty one (as it was before)

GMTA:
Thanks for the feedback @achekulaev; I've updated the PR.

GMTA:
Hi @achekulaev, I've rebased on top of the latest development commits. Are there still issues with the suggested changes?

achekulaev:
Tests were successful https://github.com/docksal/docksal/pull/1349

PR #1309, opened 2020-04-08
portainer vs poratiner

martymcflysenior:
Updated the link for portainer: [http://portainer.docksal](http://portainer.docksal/)


lmakarov:
Thanks!

PR #1353, opened 2020-05-25
Add env variables for acquia cli tool

sergey-zabolotny:


shelane:
The only problem is that the cli is expecting  ACQUIACLI_KEY and ACQUIACLI_SECRET. The secret prefix is for automatic masking. We may have to handle it like the blackfire variables for masking until we figure out a better way for the both of them.

sergey-zabolotny:
i do not understand, whats the problem with automatic masking? variables SECRET_ACQUIACLI_KEY and SECRET_ACQUIACLI_SECRET become ACQUIACLI_KEY and ACQUIACLI_SECRET respectively inside cli. did you seen this https://github.com/docksal/service-cli/pull/184?

shelane:
I’m sorry. I misunderstood then. I’m not sure what the special handling for blackfire is about.

PR #1354, opened 2020-05-25
Logging for Service Containers

sean-e-dietrich:
Added ability for all services included in stacks to have logging options configured. Additionally, added logging options of the system services.

I used defaults of 10 files and 1m file size as the default.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -6071,6 +6071,11 @@ load_global_configuration ()
 load_configuration ()
 {
 	check_project_root
+	
+	# Export Logging File Variables
+	export DOCKSAL_CONTAINER_LOG_MAX_SIZE
+	export DOCKSAL_CONTAINER_LOG_MAX_FILE
+	
 	# Re-load global env file once again, since 'up' resets some variables to force re-reading config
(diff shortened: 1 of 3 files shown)


bin/fin
```diff
@@ -1431,7 +1431,9 @@ _vhost_proxy_connect ()
 	if [[ $? == 0 ]]; then
 		echo-green "Connected vhost-proxy to \"${network}\" network."
 		# Run a dummy container to trigger docker-gen to refresh proxy configuration.
-		docker run --rm --entrypoint=echo ${IMAGE_VHOST_PROXY} >/dev/null 2>&1
```
lmakarov, line 1434: This change is not necessary. It technically does not harm, but visually it adds complexity to the code.

sean-e-dietrich: @lmakarov the function is ran when connecting containers to the vhost proxy so was thinking that it may be useful. But, if you don't think it is.

lmakarov:
Let's also add these vars to docs.

lmakarov:
@sean-e-dietrich tests are failing for this PR. Please take a look.

Issue #1352, opened 2020-05-22
Windows server IP address could not be found.


jakewhiteley-gc:
**Description**

I had a lot of issues installing so far. First I tried the virtualbox version and would get forbidden errors when accessing the *.docksal url. I decided to try the docker version instead.

No issues during install, but now the dns does not seem to be working.

When running `fin up` there are no errors and everything is reported as being successful.

I have run `fin hosts add` but still not working.

**Output of `fin config`:**

<details>
  <summary>fin config output</summary>

```
COMPOSE_PROJECT_NAME_SAFE: test
COMPOSE_FILE:
/home/jake/.docksal/stacks/volumes-bind.yml
/home/jake/.docksal/stacks/stack-default-nodb.yml
/c/Users/twent/sites/test/.docksal/docksal.yml
ENV_FILE:
/c/Users/twent/sites/test/.docksal/docksal.env

PROJECT_ROOT: /c/Users/twent/sites/test
DOCROOT: docroot
VIRTUAL_HOST: test.docksal
VIRTUAL_HOST_ALIASES: *.test.docksal
IP: 192.168.64.100
MYSQL:

Docker Compose configuration
---------------------
services:
  cli:
    dns:
    - 192.168.64.100
    - 8.8.8.8
    environment:
      BLACKFIRE_CLIENT_ID: null
      BLACKFIRE_CLIENT_TOKEN: null
      COMPOSER_ALLOW_XDEBUG: '0'
      COMPOSER_DISABLE_XDEBUG_WARN: '0'
      DOCROOT: docroot
      DRUSH_ALLOW_XDEBUG: '0'
      DRUSH_OPTIONS_URI: test.docksal
      GIT_USER_EMAIL: ''
      GIT_USER_NAME: ''
      HOST_GID: '1000'
      HOST_UID: '1000'
      MYSQL_DATABASE: default
      MYSQL_HOST: db
      MYSQL_PASSWORD: user
      MYSQL_ROOT_PASSWORD: root
      MYSQL_USER: user
      PHP_IDE_CONFIG: null
      SECRET_ACAPI_EMAIL: null
      SECRET_ACAPI_KEY: null
      SECRET_PLATFORMSH_CLI_TOKEN: null
      SECRET_SSH_PRIVATE_KEY: null
      SECRET_TERMINUS_TOKEN: null
      SSH_AUTH_SOCK: /.ssh-agent/proxy-socket
      VIRTUAL_HOST: test.docksal
[...]

PR #1355, opened 2020-05-27
Allow overriding DOCKSAL_DNS_DOMAIN with Docker Desktop 2.2.0.0+

lmakarov:
If users want to stick with .docksal on Docker Desktop on Windows, then can do so by setting manually pinning DOCKSAL_DNS_DOMAIN: "fin config set --global DOCKSAL_DNS_DOMAIN=docksal", and then use "fin hosts add project.docksal" to manage DNS records using the OS hosts file.

Relates to #1352

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -219,7 +219,8 @@ DOCKSAL_DEFAULT_DNS="8.8.8.8"
 # For visibility on this variable
 DOCKSAL_DNS_IP="${DOCKSAL_DNS_IP}"
 DOCKSAL_DNS_UPSTREAM="${DOCKSAL_DNS_UPSTREAM}"
-DOCKSAL_DNS_DOMAIN="${DOCKSAL_DNS_DOMAIN:-docksal}"
+DOCKSAL_DNS_DOMAIN_DEFAULT="docksal"
+DOCKSAL_DNS_DOMAIN="${DOCKSAL_DNS_DOMAIN:-$DOCKSAL_DNS_DOMAIN_DEFAULT}"
 # Allow disabling the DNS resolver configuration (in case there are issues with it). Set to "true" to activate.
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -7709,10 +7710,13 @@ fi
 # out of the box without the need to ask the user to manually configure DNS records using "fin hosts".
 # See https://github.com/docksal/docksal/issues/1276
 if is_wsl && is_docker_native; then
-	# Limit this to Docker Desktop for Windows 2.2.0.0+, which it is absolutely necessary
-	if (( $(ver_to_int $(docker_desktop_version)) >= $(ver_to_int '2.2.0.0') )); then
-		export DOCKSAL_NO_DNS_RESOLVER=1
-		export DOCKSAL_DNS_DOMAIN=docksal.site
+	# Only proceed if DOCKSAL_DNS_DOMAIN was not explicitly set (matches the default)
+	if [[ "${DOCKSAL_DNS_DOMAIN}" == "${DOCKSAL_DNS_DOMAIN_DEFAULT}" ]]; then
```
achekulaev, line 7714:
```suggestion
	if [[ "${DOCKSAL_DNS_DOMAIN}" == "" || "${DOCKSAL_DNS_DOMAIN}" == "${DOCKSAL_DNS_DOMAIN_DEFAULT}" ]]; then
```

lmakarov: @achekulaev can you help me understand the need for the extra condition?
`DOCKSAL_DNS_DOMAIN` can only be empty if the user sets it so. Even then, `[[ "${DOCKSAL_DNS_DOMAIN}" == "${DOCKSAL_DNS_DOMAIN_DEFAULT}" ]]` check is sufficient IMO.

achekulaev: Yeah it's not needed. Missed that case is processed above

PR #1365, opened 2020-06-13
Document how to run cron by loading bash shell

sean-e-dietrich:
Documented how to set up a crontab and how to load the bash profile so that the commands and shell are available to run.

Closes https://github.com/docksal/service-cli/issues/187

### DIFF
--- a/docs/content/service/cli/cron.md
+++ b/docs/content/service/cli/cron.md
@@ -24,7 +24,7 @@ A crontab file has six fields for specifying minute, hour, day of month, month, 
 +------------- min (0 - 59)
 ```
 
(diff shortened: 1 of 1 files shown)


docs/content/service/cli/cron.md
```diff
@@ -24,6 +24,14 @@ A crontab file has six fields for specifying minute, hour, day of month, month,
 +------------- min (0 - 59)
 ```
 
+By default cron runs commands through `/bin/sh` and can sometimes cause issues when trying to run commands that are
```
shelane, line 27: By default,

lmakarov:
Thanks!

PR #1373, opened 2020-07-07
docs mkcert addon

### LINKED ISSUE #1370: Docs: Local trusted HTTPS certs using mkcert
Document mkcert addon - https://github.com/docksal/addons/tree/master/mkcert

Demo: https://asciinema.org/a/344708

Scope:

- [x] https://docs.docksal.io/fin/addons (update - include mkcert in the list, add a section for global addons)
- [x] https://docs.docksal.io/tools/mkcert (add new)
- [ ] Write a blog post for https://blog.docksal.io/

Related:

- https://github.com/docksal/docksal/issues/498
- https://github.com/docksal/service-vhost-proxy#default-and-custom-certs-for-https

shelane:
closes #1370

### DIFF
--- a/docs/content/fin/addons.md
+++ b/docs/content/fin/addons.md
@@ -32,16 +32,24 @@ Here is a list of already created addons that can help extend functionality for 
 
 |   Name	|  Description 	|  Requirements 	|
 |--- |--- |--- |
-|   example	|  A working documentation on what is Docksal addon and how it works 	|  |
-|   adminer | [Adminer](https://www.adminer.org/) database management tool | MySQL |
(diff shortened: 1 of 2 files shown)


lmakarov:
@shelane I was thinking we document how to use mkcert (the tool) without the addon as the default option and then mention mkcert addon as the secondary option. Similar to [PHPMyAdmin](https://docs.docksal.io/tools/phpmyadmin/). This way the docs are not about the addon, but about the tool/integration and ways to use it with Docksal.

shelane:
I understand what you want now. I'll get it updated.

Issue #1268, opened 2020-01-22
Updating to Docker Desktop 2.2.0.0 breaks Docksal


ygerasimov:
I have updated my Docker Desktop to 2.2.0 and can't start docksal anymore. When I run `fin up` I am getting an error
```
Resetting Docksal system services...
 * proxy
docker: Error response from daemon: driver failed programming external connectivity on endpoint docksal-vhost-proxy (83b3ef7bafa2d27d0af784b64c635e1fff31b26803dad7bca9a4b20eb98457fb): Error starting userland proxy: listen tcp 192.168.64.100:8443: bind: cannot assign requested address.
 ERROR:  Failed starting the proxy service.
```

I did try to change ports but didn't help. My docksal version is
```
ygerasimov$ fin --version
fin version: 1.93.0
```

achekulaev:
Right now the solution is to downgrade

lmakarov:
Here's the direct link to the latest 2.1.x version (2.1.0.5):
https://download.docker.com/mac/stable/40693/Docker.dmg

lmakarov:
Upstream issue: https://github.com/docker/for-mac/issues/4209

kenrbnsn:
One gets the same error with the Windows 10 version of Docker 2.2.0.0

tekNorah:
I am also experiencing the same issue (Mac).

42robotsCS:
I can confirm this bug when upgrading to Docker Desktop 2.2.0.0 resulting in: 

`Error starting userland proxy: listen tcp 192.168.64.100:443: bind: cannot assign requested address.`

Going back to Docker Desktop Version 2.1.0.5 resolves the issue for now.

rabithk:
+1

cmodern:
+1  2

uniquename:
+1

akosglue:
+1 4

achekulaev:
I do not see any point of adding "+1" comments here. It is an upstream issue. I urge you to go and leave your comments there https://github.com/docker/for-mac/issues/4209

nerfchicken:
While it may be an upstream issue, I don't think it impacts only docker/for-mac. The same issue impacted all of our Windows users when they upgraded to 2.2.0

tekNorah:
> While it may be an upstream issue, I don't think it impacts only docker/for-mac. The same issue impacted all of our Windows users when they upgraded to 2.2.0

[...]

PR #1380, opened 2020-07-17
Unpin Docker Desktop

froboy:
As per https://github.com/docksal/docksal/issues/1268#issuecomment-628946613

Removes the Docker for Desktop warning for Mac and Windows install instructions, since that issue has been resolved. Corresponding PR on `https://github.com/docksal/docksal.io/blob/master/src/pages/installation.js` pending.

### DIFF
--- a/docs/content/getting-started/setup.md
+++ b/docs/content/getting-started/setup.md
@@ -87,14 +87,9 @@ With this method, Docker will run inside a VM in VirtualBox.
 
 ### macOS with Docker Desktop {#install-macos-docker-for-mac}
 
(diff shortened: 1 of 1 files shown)


achekulaev:
Yeah, agree with what Leonid said above. We should update the version, but it has been more than 2 times when Docker updates broke Docksal and other similar software. So installing versions above recommended is as purely at your own risk, as it is with below recommended ones.

lmakarov:
Closing via e25205d24aa05b62ae0fb4e73e981b5c567aa03e

PR #1391, opened 2020-07-21
Adjusted approach to mounting ssh-agent socket in cli

### LINKED ISSUE #1390: Invalid cross-device link with ssh-agent forwarding on Linux
When using ssh-agent forwarding in CI mode with a [docksal/sandbox-server](https://github.com/docksal/sandbox-server), the following error shows up when using `fin project` commands (anything involving docker-compose under the hood):

```
ln: failed to create hard link '/home/build-agent/.docksal/tmp/temp/agent.1909' => '/tmp/ssh-KweeHCNAIt/agent.1909': Invalid cross-device link
```

cc @lpeabody

lmakarov:
Fixes #1390

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -452,12 +452,13 @@ docker-compose ()
 		unset SSH_AUTH_SOCK
 	fi
 
-	# always create directory for socket
-	export SSH_AUTH_SOCK_DIR="$CONFIG_DIR/tmp/${COMPOSE_PROJECT_NAME_SAFE}"
-	mkdir -p "$SSH_AUTH_SOCK_DIR"
-
-	# Create original socket copy (hardlink) and update SSH_AUTH_SOCK with new socket path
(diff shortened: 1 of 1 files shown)


lmakarov:
After writing all that up, I'm thinking that we could just mount `/tmp` from host into cli and call it done.
`/tmp` is persistent on the host, so the mount will never break. ssh-agent sockets are always created in `/tmp` and we pass the actual `SSH_AUTH_SOCK` value via env variable to cli already. Need to give that a try. cc @sergey-zabolotny

lmakarov:
> I'm thinking that we could just mount `/tmp` from host into cli and call it done.

While this is definitely the easiest approach, it can lead to potential security implications. As all ssh-agent sockets are created  within `/tmp` on the host, all container that have host `/tmp` mount, will get access to all active ssh-agent sockets from the host. This is not a concern for a local environment, but does poses a risk on a shared sandbox server.

PR #1402, opened 2020-08-13
Update system-ssh-agent.md

jajajaime:
Missing "add" on the second CLI ref example.


lmakarov:
@jajajaime thanks for pointing this out. Fixed along with a few more adjustments in #1424

PR #1417, opened 2020-10-06
Added "none" stack type for projects just interested in DNS and Proxy

bioshazard:
Team, thank you first of all for Docksal. It has been immeasurably valuable to my teams. I have read [How to Contribute](https://github.com/docksal/docksal/blob/develop/CONTRIBUTING.md) and hope you will find this contribution valuable.

This simple PR introduces a "none" stack type for those that love your DNS and Vhost Proxy solution but that don't want to use the otherwise required CLI service. I tried this locally and it works exactly as I intend for a containerized Ansible development platform to replace the resource-costly VirtualBox. I am currently abusing the `node` stack type and overriding the CLI service, but would love a `none` type as is possible with `DOCKSAL_VOLUMES` to permit me to discontinue use of the `cli`-named service.

It may feel simply decorative, but there have been a few cases in my development where a "default CLI container" didn't feel right and we wind up setting `cli` to use a default alpine image so it will just auto-exit bash on project start.

### DIFF
--- a/stacks/stack-none.yml
+++ b/stacks/stack-none.yml
@@ -1,0 +1,5 @@
+# Blank stack
+#
+# Useful for custom docker-compose needs that want to use Docksal DNS and Proxy
+
+version: "2.1"


lmakarov:
@bioshazard glad you and your team are enjoying Docksal!

We've been discussing `DOCKSAL_STACK=none` multiple times before, so thanks for bringing this up =)

Just to confirm, with the proposed `none` option for stack, you would still be using `.docksal/docksal.yml` to put together a custom docker-compose configuration for the project stack, correct?

bioshazard:
@lmakarov correct. Right now I am using `.docksal/docksal.yml` and specifying the `cli` image as `alpine` so it will just exit on start. I'd love to be able to use a `none` stack to exclude it entirely. I tested this locally and it yields the exact result I want. Even outside of my likely abusive ansible solution, I have experience with a highly couple publish/subscribe solution that side-stepped the use of the `cli` service in the same way.

bioshazard:
Oh I see now the PR you linked. This PR I have provided would eliminate the need for a condition. But either solution would deliver my interest. No hard feelings if this PR gets eliminated in favor of the condition being added to `fin`.

lmakarov:
@achekulaev and I discussed this use case again and it's actually already fully supported and only needs proper documentation.

You simply don't set the `DOCKSAL_STACK` variable in `.docksal/docksal.env` and put your custom docker-compose stack definition in `.docksal/docksal.yml`.

If you also did not want any of the managed volumes injected into the stack, then set `DOCKSAL_VOLUMES=disable` (NOTE: this will be changed to `DOCKSAL_VOLUMES=disabled` in the next release).

Here's what the end result would look like:

```
$ cat ./.docksal/docksal.env
DOCKSAL_VOLUMES="disabled"

 $ cat ./.docksal/docksal.yml
version: "2.1"

$ fin config
---------------------
COMPOSE_PROJECT_NAME_SAFE: test
COMPOSE_FILE:
/Users/leonid/Work/Projects/test/.docksal/docksal.yml
ENV_FILE:
/Users/leonid/Work/Projects/test/.docksal/docksal.env

PROJECT_ROOT: /Users/leonid/Work/Projects/test
DOCROOT: docroot
VIRTUAL_HOST: test.docksal
VIRTUAL_HOST_ALIASES: *.test.docksal
IP: 192.168.64.100
MYSQL:

Docker Compose configuration
---------------------
services: {}
version: '2.1'

---------------------
```

☝️ a clean state.

bioshazard:
Well ok then lol. I can confirm it works exactly as described. Thanks for the insight!

PR #1423, opened 2020-10-19
Fix Dockerfile example to use proper shell

vitaly-malashevsky:
Let example use bash as default shell as long as docksal relies on it everywhere

If it's not a bash then it produces errors like this on container creation phase:
image

### DIFF
--- a/docs/content/stack/extend-images.md
+++ b/docs/content/stack/extend-images.md
@@ -66,6 +66,7 @@ RUN \
 
 # All further commands will be performed as the docker user.
 USER docker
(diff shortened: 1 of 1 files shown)


lpeabody:
Makes sense to me. @lmakarov thoughts?

lmakarov (approved):

lmakarov:
Yep, this aligns with what we are already doing in docksal/cli:
https://github.com/docksal/service-cli/blob/develop/7.4/Dockerfile#L262

Thanks!

PR #1446, opened 2020-12-17
Create stack-platformsh.yml

waako:
Create a docksal stack alike the Platform.sh stack, based on their recommended defaults in [platformsh-template for Drupal 9](https://github.com/platformsh-templates/drupal9).

They use MariaDB 10.4, but using 10.3 here until https://github.com/docksal/service-mariadb/pull/9 is merged

We try to keep our project environments up to date with platformsh's templates, so it seems to make sense there is a stack to use rather than overriding services and images each time.

### DIFF
--- a/stacks/stack-platformsh.yml
+++ b/stacks/stack-platformsh.yml
@@ -33,7 +33,7 @@ services:
     # Pin PHP version
     image: ${CLI_IMAGE:=docksal/cli:stable-php7.4}
 
-redis:
+  redis:
     extends:
       file: ${HOME}/.docksal/stacks/services.yml
       service: redis
--- a/stacks/stack-platformsh.yml
+++ b/stacks/stack-platformsh.yml
@@ -1,0 +1,41 @@
+# Platform.sh alike stack
+# see https://github.com/platformsh-templates/drupal9
+#
+# - Nginx 1.14
+# - MariaDB 10.3
+# - PHP 7.4
+# - Redis 6.0
+
+version: "2.1"
+
+services:
+  # http(s)://VIRTUAL_HOST
+  web:
+    extends:
+      file: ${HOME}/.docksal/stacks/services.yml
+      service: nginx
+    # Pin nginx version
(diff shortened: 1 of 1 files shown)


lpeabody:
@waako agreed. I just did some Platform.sh work and this would have been handy! I was planning on getting the MariaDB and Solr images brought up-to-date after the holiday break. How do you feel about including Solr in the stack? It seems like its a fairly common inclusion in Platform.sh applications.

waako:
@lpeabody Undecided about Solr, we currently have it commented out in our "template" `docksal.yml` so that it can be enabled easily but a number of our projects don't require it, so seems reasonable to limit resources used.

Once it's in there, you can't disable it for a project (I assume), that's the main downside I see.

Maybe better to have a `stack-platformsh-solr.yml` that extends `stack-platformsh.yml` to include `solr`?

lpeabody:
We'll leave it out then. We'll just make sure that we add a reference to the documentation about adding optional Docksal services (solr included) to projects.

stacks/stack-platformsh.yml
```diff
@@ -0,0 +1,41 @@
...
+    image: ${DB_IMAGE:-docksal/mariadb:10.3-1.1}
+
+  cli:
+    extends:
+      file: ${HOME}/.docksal/stacks/services.yml
+      service: cli
+    # Pin PHP version
+    image: ${CLI_IMAGE:=docksal/cli:stable-php7.4}
+
+redis:
```
lpeabody, line 36: Needs to be indented and aligned with the rest of the service entries.

lpeabody (changes requested):

stacks/stack-platformsh.yml
waako: Oups, sorry about that, should be fixed now

waako:
@lpeabody Any ideas on when this might get merged, anything outstanding?

lpeabody:
@waako Nope, nothing outstanding, just me losing track of all the things :) Thanks for the reminder.

PR #1454, opened 2021-01-04
Make healthcheck interval configurable

### LINKED ISSUE #1453: Reduce idle load by increasing container healthcheck interval
We use container healthchecks in Docksal system and project containers.
The interval between healthchecks is configured in Dockerfile of each container image. The default value we use is **5s**.

Unfortunately, as the number of running containers on a host grows, so grows the load from all of the individual 
 heathchecks in each container. This becomes noticeable on host machines running multiple projects and containers concurrently. Even when projects sit idle, Docker constantly runs those healthchecks.

Healthcheck options can be overridden at runtime (`docker run ...` or `docker-compose.yml`). So we don't have to push updates to all container images in the Docksal library, but set the interval globally instead.

Increasing the interval from 5s to 10s can help save some CPU cycles. Laptops users on batteries will especially appreciated this change.

lmakarov:
- Added `DOCKSAL_CONTAINER_HEALTHCHECK_INTERVAL` variable
  - Default: `10s` (was `5s`, set in individual container image Docketfile's)
  - Can be used to increase container healthcheck interval and thus help reduce idle load produced by healthchecks

Resolves #1453

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -279,7 +279,8 @@ else
 	DOCKSAL_SSH_AGENT_USE_HOST="${DOCKSAL_SSH_AGENT_USE_HOST:-0}"
 fi
 
-# Healthcheck timeout
+# Project healthcheck timeout
+# Total time (seconds) before the _project_ healthcheck is considered failed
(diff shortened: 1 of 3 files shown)


lpeabody (approved):
My sys76 laptop appreciates this very much.

Issue #1448, opened 2020-12-18
fin run-cli composer uses composer 1.10.7 in fin 1.102.0


davekopecek:
The run-cli command in v1.102.0 uses composer 1.10.7, composer has been updated to 2.0.6 for `fin composer`

**Steps to reproduce the issue:**

1. Run `fin composer -V`  
2. Run `fin run-cli composer -V`

**Describe the results you received:**

```
docksal@DocksalNode ~/sandbox/test
master└─ $ fin composer -V
Composer version 2.0.6 2020-11-07 11:21:17
docksal@DocksalNode ~/sandbox/test
master└─ $ fin run-cli composer -V
Composer version 1.10.7 2020-06-03 10:03:56
```

**Describe the results you expected:**

Expected that `fin run-cli composer` would use Composer 2.x version.

shelane:
The `run-cli` is still set to 2.11 and Composer 2 was added in 2.12.

lmakarov:
Fixed in #1463

PR #1463, opened 2021-02-12
Updating cli version to latest

shelane:
The updated run-cli is necessary to have access now to Composer 2, which more projects are now updated to require.

The basic service cli update was missed with the last update.


shelane:
Fixes #1448

lmakarov (approved):

PR #1465, opened 2021-02-13
Added option for addons and other files to source the fin file

sean-e-dietrich:
Added option so that `fin` could be sourced for addons and commands and take advantage of the functions already provided in the main file.

Example of the usage:

```bash
# Find the fin command
fin=`which fin`

# Source the file to use functions
[[ -f ${fin} ]] && source $fin
```

### DIFF
--- a/docs/content/fin/custom-commands.md
+++ b/docs/content/fin/custom-commands.md
@@ -199,6 +199,20 @@ Commands are ran in the same exact way as normal except include the folder they 
 fin drupal/updb
 ```
 
+## Reusing fin functions
+
+To reuse some of the functionality within the `fin` command like `parse_params`, `is_linux`, or any of the functions it's possible to source the `fin` file.
+
+Including this in your command or addon would look like the following.
+
+```bash
+# Find the fin command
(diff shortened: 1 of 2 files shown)


lmakarov (changes requested):
Look's harmless to me. Why not.

@sean-e-dietrich please also add a note in the docs about this use case.

docs/content/fin/custom-commands.md
```diff
@@ -199,6 +199,20 @@ Commands are ran in the same exact way as normal except include the folder they
 fin drupal/updb
 ```
 
+## Reusing fin functions
+
+To reuse some of the functionality within the `fin` command like `parse_params`, `is_linux`, or any of the functions it's possible to source the `fin` file.
```
shelane, line 204: It is possible to source the `fin` file in order to reuse some of the functionality within the `fin` command like `parse_params`, `is_linux`, or any of the functions.

shelane (approved):

achekulaev (approved):

lmakarov (approved):

Issue #701, opened 2018-07-11
Move to using Compose file version 2.4


achekulaev:
Update: Since docker-compose 3.x does not support `extends` (which we widely use today in our docker-compose yamls), we should move to the latest 2.x version - v2.4 at the moment.

---

Considerations, pros and cons of moving to using Compose file version 3. One of the reasons is that it would allow using `docker stack` instead of `docker-compose`, thus removing a dependency.

sean-e-dietrich:
I don't have knowledge of `docker stack` other then what is outlined on the documentation. But, if it removes yet another dependency I'm for it.

Do we lose functionality that we currently have by upgrading?

lmakarov:
We may want to consider moving to 2.4 (brought up in #895) before even thinking about the 3.x jump.

With the exception of hardcoded `version` values in all existing project yamls, the 2.1 => 2.4 upgrade should be pretty straightforward. The 3.x upgrade won't be that easy or even doable for the time being.

Let's brainstorm ideas here around handling the 2.4 upgrade in a way that minimizes a breaking change burned on our users. This same approach can be then applied later on with the 3.x shift.

achekulaev:
Oh!... I did not check that... Well the yeah we should update to the latest 2.x

PR #1479, opened 2021-03-20
Switch to Docker Compose file v3

lmakarov:
v2 and v3 Docker Compose file formats were [merged together](https://github.com/docker/compose/pull/7588) in docker-compose v1.27+

This means, we can safely switch to the latest compose file v3.9 and continue to use `extends` + all of the new features and options that we were not able to use while stuck on 2.1, like [extension-fields](https://docs.docker.com/compose/compose-file/compose-file-v3/#extension-fields) and [yaml anchors](https://yaml.org/spec/1.2/spec.html#id2765878).

Moreover, the "version" property is now optional and combining files with different versions will not result in an error. This used to be a major blocker for us to even consider updating the compose file format version, since that would be a breaking change for every single Docksal project out there.

Closes: #701

### DIFF
--- a/stacks/services.yml
+++ b/stacks/services.yml
@@ -35,6 +35,9 @@ services:
       - APACHE_FCGI_HOST_PORT=cli:9000
       - APACHE_BASIC_AUTH_USER
       - APACHE_BASIC_AUTH_PASS
+    << : *dns
+    << : *logging
+    << : *healthcheck
 
   # Web: Nginx
   nginx:
--- a/stacks/services.yml
+++ b/stacks/services.yml
@@ -4,6 +4,20 @@
 
 version: "3.9"
 
+x-common-settings:
+  dns: &dns
+    dns:
+      - ${DOCKSAL_DNS1}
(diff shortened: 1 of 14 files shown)


stacks/services.yml
```diff
@@ -21,15 +35,6 @@ services:
...
       - APACHE_BASIC_AUTH_PASS
-    dns:
-      - ${DOCKSAL_DNS1}
-      - ${DOCKSAL_DNS2}
-    logging:
-      options:
-        max-size: ${DOCKSAL_CONTAINER_LOG_MAX_SIZE}
-        max-file: ${DOCKSAL_CONTAINER_LOG_MAX_FILE}
-    healthcheck:
-      interval: ${DOCKSAL_CONTAINER_HEALTHCHECK_INTERVAL}
```
lpeabody, line 32: Are these meant to be replaced also by their respective anchor references?

lmakarov: Oops, missed that one. Thanks.

lpeabody (approved):

PR #1481, opened 2021-03-23
Update acquia-cli.md

acouch:
Updates acquia cli docs.

Not sure about: 

```
Use `fin acli <command>` from the host or `acli <command>` inside `cli`.`
```

Looks like the top level command doesn't exist?

### DIFF
--- a/docs/content/tools/acquia-cli.md
+++ b/docs/content/tools/acquia-cli.md
@@ -36,7 +36,7 @@ As such, it is best to use `.docksal/docksal-local.env` for any personal keys an
 
 ## Usage
 
-Use `fin acli <command>` from the host or `acli <command>` inside `cli`.
+Use `fin exec acli <command>` from the host or `acli <command>` inside `cli`.
 
 Please refer to the [Acquia CLI readme](https://github.com/acquia/cli) for usage details and
 the [Acquia documentation](https://docs.acquia.com/acquia-cloud/develop/api/) for the options available
--- a/docs/content/tools/acquia-cli.md
+++ b/docs/content/tools/acquia-cli.md
@@ -8,7 +8,7 @@ aliases:
 
 ## Using Acquia Cloud CLI Commands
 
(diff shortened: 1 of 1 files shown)


shelane:
`fin exec acli`

shelane (changes requested):
Lines 11-14 also need to be updated. It now requires 2.13.0+

shelane (approved):

lpeabody (approved):

lmakarov (changes requested):
We have `terminus` and `platform` as top-level commands.
Let's be fair and treat `acli` the same way. It's only a few lines of code :)

@acouch see https://github.com/docksal/docksal/blob/a8af03696a5f2013e04936ce314375fd97159325/bin/fin#L8386-L8390 for an example.

lmakarov:
ping @acouch. Will you be adding the top level `fin acli` command as part of this PR?

docs/content/tools/acquia-cli.md
```diff
@@ -22,8 +22,8 @@ generate your Acquia Cloud API Token.
 Add your secret and key with
 
 ```bash
-fin config set --global SECRET_ACQUIACLI_SECRET="lkjd*sdfl2//23498fwernoiuDHljw3897fsk"
-fin config set --global SECRET_ACQUIACLI_KEY="rSxVZN35bo4jTuncGS+CiKdmhxLPL0BaPuyOv"
+fin config set --global SECRET_ACQUIA_CLI_SECRET="lkjd*sdfl2//23498fwernoiuDHljw3897fsk"
+fin config set --global SECRET_ACQUIA_CLI_KEY="rSxVZN35bo4jTuncGS+CiKdmhxLPL0BaPuyOv"
```
shelane, line 26: These probably should not be changed because `SECRET_ACQUIACLI_SECRET` and `SECRET_ACQUIACLI_KEY` are what are passed in automatically to the CLI and many users may already be using the existing variables. There is no functional reason to change them.

shelane (changes requested):

PR #1486, opened 2021-04-15
adds warning for changing volume mount type

### LINKED ISSUE #1461: DOCKSAL_VOLUMES incorrect documentation
**Description**

Incorrect documentation in [Shared Volumes / DOCKSAL_VOLUMES](https://docs.docksal.io/core/volumes/#docksal-volumes). It reports in case of DOCKSAL_VOLUMES switch: 

_Once you set a new DOCKSAL_VOLUMES option, you must recreate the cli container. The easiest way is fin project reset, but it will also remove all data from the db volume. If you want to retain it, remove only the cli container, and start the project again: fin project remove cli; fin project start_

but `fin project remove cli && fin project start` won't work because project_root volume is used in all containers (for all cases except cli as read-only) so it will produce for a test project: 

`ERROR: Configuration for volume project_root specifies "o" driver_opt None, but a volume with the same name uses a different "o" driver_opt (addr=192.168.64.1,vers=3,nolock,noacl,nocto,noatime,nodiratime,actimeo=1). If you wish to use the new configuration, please remove the existing volume "test_project_root" first:
$ docker volume rm test_project_root`

I add a question, is project_root volume needed in read-only mode for all containers except for cli? Maybe can it be used in cli and web only? In this way with `fin project remove cli && fin project remove web && docker volume rm myproj_project_root && fin restart` we can achieve the result without a `fin project reset` which will remove db data too.

shelane:
Add warning to core volumes section on changing volume mount type.

closes #1461

### DIFF
--- a/docs/content/core/volumes.md
+++ b/docs/content/core/volumes.md
@@ -185,10 +185,9 @@ fin project reset
 
 Wait until the initial unison sync finishes.
 
-{{% notice note %}}
-Once you set a new `DOCKSAL_VOLUMES` option, you must recreate the `cli` container. The easiest way is `fin project reset`,
-but it will also remove all data from the `db` volume. If you want to retain it, remove only the `cli` container, and 
-start the project again: `fin project remove cli; fin project start`
+{{% notice warning %}}
+Switching volumes is a destructive operation for the whole project stack. To preserve the current state of the database,
+it has to be exported before it is changed and then imported back.
 {{% /notice %}}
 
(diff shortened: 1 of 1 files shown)


shelane:
Here is the relevant page change: https://deploy-preview-1486--docs-docksal-io.netlify.app/core/volumes/#docksal-volumes

lmakarov (changes requested):
Also need the note in this section updated (see #1461 description)
https://deploy-preview-1486--docs-docksal-io.netlify.app/core/volumes/#using-unison-volumes

lmakarov (approved):

PR #1490, opened 2021-05-05
Make headless browsers available.

lpeabody:
# Changes

- Makes the following SeleniumHQ-supported headless browsers available for usage within Docksal:
    - Chrome
    - Firefox
    - Edge
- Browsers can be CPU intensive, so they are disabled by default (even if you've extended them into your `docksal.yml` file). You must explicitly add `COMPOSE_PROFILES="testing"` to either `docksal.env` or `docksal-local.env` and then run `fin up`.

### DIFF
--- a/docs/content/service/other/browsers.md
+++ b/docs/content/service/other/browsers.md
@@ -16,7 +16,7 @@ browser images:
 To include one or more of these browsers with your project, update the 
 `services` section of `.docksal/docksal.yml` by adding:
 
-*Chrome*
+**Chrome**
 
 ```yaml
   chrome:
@@ -25,7 +25,7 @@ To include one or more of these browsers with your project, update the
       service: chrome-browser
 ```
 
-*Firefox*
+**Firefox**
 
 ```yaml
   firefox:
@@ -34,7 +34,7 @@ To include one or more of these browsers with your project, update the
       service: firefox-browser
 ```
 
-*Edge*
+**Edge**
 
 ```yaml
   edge:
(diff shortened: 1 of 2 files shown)


lpeabody:
The chrome browser presents some issues. You may receive a couple of errors immediately:

Tab crash (error is generated from PHPUnit test):

```
...
WebDriver\Exception\UnknownError: unknown error: session deleted because of page crash
from unknown error: cannot determine loading status
from tab crashed
  (Session info: chrome=84.0.4147.105)
...
```

Blocked connection (Docker logs from browser service):

```
...
browser_1    | Only local connections are allowed.
browser_1    | Please see https://chromedriver.chromium.org/security-considerations for suggestions on keeping ChromeDriver safe.
browser_1    | [Ch1romeD6r2i0v2e3r0 8w1a6s. 5s8t4a]r[tSeEdV EsRuEc]c:e sbsifnudl(l)y .f
browser_1    | ailed: Cannot assign requested address (99)
...
```

Both of the errors above, based on my testing locally, can be resolved by updating the browser service definition like so:

```
browser:
    hostname: browser
    image: selenium/standalone-chrome
    environment:
      JAVA_OPTS: "-Dwebdriver.chrome.whitelistedIps="
    shm_size: '512M'
    dns:
      - ${DOCKSAL_DNS1}
      - ${DOCKSAL_DNS2}
```

Setting `shm_size` addresses the tab crash, and adding the JAVA_OPTS whitelist equal to nothing allows all remote connections. If we wanted to be more secure we could make the browser service dependent on the cli service and figure out a way to pass the IP address of the cli service on the project network.

For more details regarding the `shm_size` property usage in the context of the selenium-chrome image, see https://stackoverflow.com/a/53970825/664881.

PR #1493, opened 2021-05-14
adds check for existing host name before adding to eliminate duplicates

shelane:
Adding a grep check of hostname in hosts file so that if `fin hosts add` is called multiple times, the hosts file won't contain multiple duplicates.

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -6705,22 +6705,26 @@ hosts_add ()
 	fi
 	[[ ! "$hostname" =~ "." ]] && echo-error "Hostname should contain '.'" && exit 1
 
-	# Always make hosts file backup
-	hosts_backup
+  if ! (cat "$HOSTS_FILE" | grep -e "$DOCKSAL_IP $hostname" >/dev/null); then
+    # Always make hosts file backup
+    hosts_backup
 
-	# Add to hosts
-	echo -e "Adding: ${yellow}$DOCKSAL_IP $hostname${NC}"
-	echo -e "Adding: ${yellow}::2 $hostname${NC}"
-	if is_windows; then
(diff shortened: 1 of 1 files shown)


bin/fin
```diff
@@ -6705,22 +6705,26 @@ hosts_add ()
...
-		local tmp="/tmp/hosts.bak.$RANDOM"
-		cp "$HOSTS_FILE" "$tmp"
-		echo "$DOCKSAL_IP $hostname" | tee -a "$tmp" >/dev/null
-		echo "::2 $hostname" | tee -a "$tmp" >/dev/null
-		unix2dos -q "$tmp"
-		winsudo "copy /Y $(cygpath -w ${tmp}) $(cygpath -w ${HOSTS_FILE}) && erase $(cygpath -w ${tmp})"
-	else
-		echo "$DOCKSAL_IP $hostname" | sudo tee -a "$HOSTS_FILE" >/dev/null
-		echo "::2 $hostname" | sudo tee -a "$HOSTS_FILE" >/dev/null
+  if ! (cat "$HOSTS_FILE" | grep -e "$hostname" >/dev/null); then
```
lpeabody, line 6708: I think this results in partial matches. Like if host name was "thissite.docksal" and there was already an entry in hosts for "subdomain.thissite.docksal", this would fail and result in thissite.docksal not being added even though its not actually present.

lpeabody (changes requested):
Just one comment on the method for checking the host name.

This might also be worthy of a BATS test.

shelane:
I have changed the check to exactly match the entry to remove the risk of finding the domain from an existing subdomain entry. I don't know anything about BATS tests though.

lmakarov (approved):

PR #1500, opened 2021-06-16
Version bumps

lmakarov:
- Docker Desktop v3.4.0 (65384)
- Docker v20.10.7
- Docker Compose v1.29.2
- VirtualBox v6.1.22


lmakarov:
Had to implement a hack in fin (55bbedb) to ignore docker version checks when running in GitHub Actions.

The problem there is that the get.docker.com script we use to automatically install/update a specific docker engine version does not properly work in GitHub Actions. We'll have to rely on the preinstalled version of docker in the GitHub Actions ubuntu runner that we use (`ubuntu-20.04` a the moment).

PR #1503, opened 2021-06-17
Docker Compose v2

### LINKED ISSUE #1502: Switch to docker compose (compose v2)
The new `docker compose` ([compose v2](https://docs.docker.com/compose/cli-command/)) command is a drop-in replacement for docker-compose v1:

```
❯ docker compose

Usage:  docker compose [OPTIONS] COMMAND

Docker Compose

Options:
      --ansi string                Control when to print ANSI control characters ("never"|"always"|"auto") (default "auto")
      --env-file string            Specify an alternate environment file.
  -f, --file stringArray           Compose configuration files
      --profile stringArray        Specify a profile to enable
      --project-directory string   Specify an alternate working directory
                                   (default: the path of the Compose file)
  -p, --project-name string        Project name

Commands:
  build       Build or rebuild services
  convert     Converts the compose file to platform's canonical format
  cp          Copy files/folders between a service container and the local filesystem
  create      Creates containers for a service.
  down        Stop and remove containers, networks
  events      Receive real time events from containers.
  exec        Execute a command in a running container.
  images      List images used by the created containers
  kill        Force stop service containers.
  logs        View output from containers
  ls          List running compose projects
  pause       pause services
  port        Print the public port for a port binding.
  ps          List containers
  pull        Pull service images
  push        Push service images
  restart     Restart containers
  rm          Removes stopped service containers
  run         Run a one-off command on a service.
  start       Start services
  stop        Stop services
  top         Display the running processes
  unpause     unpause services
  up          Create and start containers

Run 'docker compose COMMAND --help' for more information on a command.
```

Unlike docker-compose v1, compose v2 is available for ARM64 (Linux and Mac).

lmakarov:
- Installing arch specific docker binaries
  - Added `host_arch` (`x86_64` / `aarch64`) and `host_arch_new` (`amd64` / `arm64`) functions to read host architecture
  - Updated download URLs for docker binaries to include host arch
- Switched to Docker Compose v2
  - Compose v2 binaries are available for both `amd64` and `arm64` architectures.
  - Should be a drop-in replacement for docker-compose v1 (with the exception of how the binary is invoked)
  - Set `COMPOSE_COMPATIBILITY=true` to force underscores instead of hyphens in container names (like in v1).
  - Fixed containers not connected to project network after stack restart 
    - While Docker Compose v2 recreates the project network if one does not exist, it does not reconnect project containers to the network. We have to do this manually.
- Updated `sysinfo` output for docker and docker compose
- Updated tests

Fixes #1502


bin/fin
```diff
@@ -115,10 +115,29 @@ is_katacoda ()
...
+		x86_64) echo "x86_64" ;;
+		aarch64) echo "aarch64" ;;
+		* ) false;;
+	esac
+}
+
+# Returns amd64 / arm64 (aka "new style")
+host_arch_new () {
+	case "$(uname -m)" in
+		x86_64) echo "amd64" ;;
```
lpeabody, line 135: @lmakarov should `x86_64` here be a different value?

lpeabody (changes requested):
I think I caught one logic condition error.

bin/fin
lmakarov: @lpeabody `uname -m` returns `x86_64` or `aarch64`.
Both `host_arch` and `host_arch_new` function use it but return different architecture styles.

- `x86_64` = `amd64`
- `aarch64` =  `arm64`

lpeabody: Ah okay that clears up some confusion I had.

lmakarov:
@lpeabody do you have any other feedback?

lmakarov:
ping @lpeabody, @achekulaev, @sean-e-dietrich

bin/fin
```diff
@@ -119,10 +119,29 @@ is_katacoda ()
...
+host_arch () {
+	case "$(uname -m)" in
+		x86_64) echo "x86_64" ;;
+		aarch64) echo "aarch64" ;;
+		* ) false;;
+	esac
+}
+
+# Returns amd64 / arm64 (aka "new style")
+host_arch_new () {
```
achekulaev, line 137: The function is not used anywhere, can be deleted

bin/fin
```diff
@@ -1313,11 +1373,14 @@ _start_containers ()
 	check_docker_path
 
 	echo-green "Starting services..."
-	docker-compose up -d --remove-orphans --build || return 1
+	docker-compose up -d --remove-orphans --build --quiet-pull || return 1
```
achekulaev, line 1376: Pull can take many minutes when install is fresh. Progress was providing valuable visual feedback. --quiet-pull will remove that

bin/fin
```diff
@@ -4525,10 +4593,11 @@ install_tools_debian ()
 
 	# Install Docker Compose
 	if ! is_docker_compose_version; then
-		echo-green "Installing Docker Compose..."
+		local dc_pkg=$(basename "$URL_DOCKER_COMPOSE_NIX")
+		echo-green "Installing docker-compose v${REQUIREMENTS_DOCKER_COMPOSE}..."
```
achekulaev, line 4597: Not a critical thing, but I guess we should always call it "downloading $toolname" rather than "installing $toolname" to prevent freaking people out

bin/fin
```diff
@@ -151,6 +170,19 @@ wsl_path ()
...
 }
 
+# Convert version string like 1.2.3.4 to integer for comparison
+# param $1 version string of 4 components max (e.g., 1.10.3.0)
+ver_to_int ()
+{
+	echo "$@" | awk -F. '{ printf("%d%03d%03d%03d", $1,$2,$3,$4); }'
+}
+
+# docker-compose v1.x.x vs docker compose v2.x.x
```
achekulaev, line 180: I would advocate to only support dc v 2.x and get rid of everything 1.x related

tests/project.bats
```diff
@@ -27,63 +27,42 @@ teardown() {
...
 
 @test "fin project stop" {
 	[[ $SKIP == 1 ]] && skip
 
-	run fin project stop
-	echo "$output" | egrep "Stopping .*_web_1"
-	echo "$output" | egrep "Stopping .*_db_1"
-	echo "$output" | egrep "Stopping .*_cli_1"
-	unset output
+	fin project stop
```
achekulaev, line 43: Seems like all the tests from here and below has been castrated, and do not perform any real tests that things are going as they should, but rather just check for the exit code. Just throwing in concerns here that it might not provide adequate testing of fin now.

bin/fin
lmakarov: This was in use for Docker Compose v2-beta download URLs, then they switched to the other variant for the v2 release. There does not seem to be a single standard and we may need the "new style" version at some point with other tools. I'd rather have it there for completeness.

bin/fin
lmakarov: Yeah, keeping v1 made sense during v2-beta for testing and comparison. With the v2 release and all tests passing, we can cleanup the v1 stuff.

tests/project.bats
lmakarov: Fair point.
Compose v1 and v2 generate different intermediary outputs. Accounting for both would bloat the tests cases.
If we drop everything around the v1 implementation, then this would not be an issue.

bin/fin
lmakarov: While it looks nice in an interactive shell, pull produces a log of garbage in the CI logs.

Interactive shell:

Screen Shot 2021-11-09 at 08 27 26

Ci logs:

Screen Shot 2021-11-09 at 08 29 02

With `--quiet-pull` user still gets a "spinner" instead of the full progress output:

Screen Shot 2021-11-09 at 08 26 23

@achekulaev  If you feel like the humble pull spinner output is not sufficient, then we can do `--quiet-pull` only for non-interactive shells (`is_tty != 0`).

Note: CIs that fake TTY (I know Travis does this, not sure about the rest) would still suffer from the enormous pull output.

lmakarov:
@achekulaev thanks for your feedback. I'm going to update the PR and resubmit for review.

bin/fin
```diff
@@ -355,18 +387,27 @@ STACKS_VOLUMES_NONE="volumes-none.yml"
...
-URL_DOCKER_COMPOSE_NIX="https://github.com/docker/compose/releases/download/${REQUIREMENTS_DOCKER_COMPOSE}/docker-compose-Linux-x86_64"
-URL_DOCKER_COMPOSE_WIN="https://github.com/docker/compose/releases/download/${REQUIREMENTS_DOCKER_COMPOSE}/docker-compose-Windows-x86_64.exe"
-URL_DOCKER_MACHINE_MAC="https://github.com/docker/machine/releases/download/v${REQUIREMENTS_DOCKER_MACHINE}/docker-machine-Darwin-x86_64"
-URL_DOCKER_MACHINE_NIX="https://github.com/docker/machine/releases/download/v${REQUIREMENTS_DOCKER_MACHINE}/docker-machine-Linux-x86_64"
-URL_DOCKER_MACHINE_WIN="https://github.com/docker/machine/releases/download/v${REQUIREMENTS_DOCKER_MACHINE}/docker-machine-Windows-x86_64.exe"
+# docker-compose v1
+URL_DOCKER_COMPOSE_MAC="https://github.com/docker/compose/releases/download/${REQUIREMENTS_DOCKER_COMPOSE}/docker-compose-Darwin-$(host_arch)"
+URL_DOCKER_COMPOSE_NIX="https://github.com/docker/compose/releases/download/${REQUIREMENTS_DOCKER_COMPOSE}/docker-compose-Linux-$(host_arch)"
+URL_DOCKER_COMPOSE_WIN="https://github.com/docker/compose/releases/download/${REQUIREMENTS_DOCKER_COMPOSE}/docker-compose-Windows-$(host_arch).exe"
```
sean-e-dietrich, line 399: Do we still need these links if we are forcing requirements to be v2?

bin/fin
```diff
@@ -520,9 +561,19 @@ docker-compose ()
...
+		# As a plugin for docker cli, it has to be called as "/path/to/binary compose"
+		# This would not be necessary if the binary was installed in the default location
+		# (~/.docker/cli-plugins/docker-compose). However, we install it in ~/.docksal/bin/docker-copose-v2
+		# to prevent conflicts with system binaries.
+		# "command <binary>" calls the binary directly ignoring any overriding shell functions
+		command docker-compose compose "$@"
+	else
+		# Call regular docker-compose.
+		# "command <binary>" calls the binary directly ignoring any overriding shell functions
+		command docker-compose "$@"
```
sean-e-dietrich, line 575: I guess again do we need this if we are forcing them to use v2?

bin/fin
sean-e-dietrich: I am a fan of seeing the download piece to see the true progression of everything.

bin/fin
sean-e-dietrich: I think i left similar comments that if we are forcing the v2 on the user we could remove any conditionals around checking for v1

bin/fin
lmakarov: Switched to doing `--quiet-pull` only for CI environments.

bin/fin
lmakarov: Removed all references and support for Docker Compose v1

bin/fin
lmakarov: Removed all references and support for Docker Compose v1

tests/project.bats
lmakarov: Updated `fin project` tests for Docker Compose v2

bin/fin
lmakarov: Removed all references and support for Docker Compose v1

bin/fin
lmakarov: Updated wording

lmakarov:
@achekulaev, @sean-e-dietrich made all of the updates. Tests are green. Eager to get this out of the door 😃.

bin/fin
```diff
@@ -119,10 +119,33 @@ is_katacoda ()
...
 }
 
+# Returns x86_64 / aarch64 (aka "old school")
+host_arch () {
+	case "$(uname -m)" in
+		x86_64) echo "x86_64" ;;
+		amd64) echo "x86_64" ;;
+		aarch64) echo "aarch64" ;;
+		arm64) echo "aarch64" ;;
+		* ) false;;
```
achekulaev, line 134: This case will lead to issues in URLs. We should default to something e.g. x86_64

bin/fin
```diff
@@ -119,10 +119,33 @@ is_katacoda ()
...
+	esac
+}
```
achekulaev, line 136:
```suggestion
host_arch () {
    case "$(uname -m)" in
        aarch64|arm64) echo "aarch64";;
        *) echo "x86_64";;
    esac
}
```

lmakarov: If the host is neither x86_64, nor aarch64, then defaulting to x86_64 would not help as nothing else will work anyway. I'd rather have this function return `false` if host arch is not expected and add a check at the beginning of runtime:

```bash
if ! host_arch; then 
    echo "Host architecture not supported."
    exit 1
fi
```

sean-e-dietrich (approved):

bin/fin
lmakarov: @achekulaev I added the host architecture check and also made it so that `host_arch`/`host_arch_new` always return a string value (returning `uname -m` in the catch-all case).

AlexNetman:
Huge!

PR #1508, opened 2021-06-25
Issue 1507: Adding variable for 3.x wodby/xhprof.

### LINKED ISSUE #1507: Update XHProf service to work with XHProf 2.x and 3.x
**Description**

It seems that the 3.x version of XHProf was updated to use the variable `XHPROF_OUTPUT_DIR` instead of `PHP_XHPROF_OUTPUT_DIR`.  This led to the default UI of the container that lives at `http://xhprof.[sitename].docksal` to not have access to any of the runs.

_Note: Requires a vanilla Drupal setup to compare_
1. Install [XHProf](https://drupal.org/projects/xhprof) module on Drupal site.
2. Run `fin config set --env=local XHPROF_ENABLED=1`
3. Run `fin up` -- This should pull and install the wodby/xhprof container
4. Configure XHProf module in Drupal
5. Profile a page
6. View profile in Drupal `/admin/reports/xhprof`
7. Visit http://xhprof.[sitename].docksal
8. Note that no results are showing
3. 

No runs showing

Matching runs showing

This appears to be fixed by adding the new variable to the `overrides-xhprof.yml` file, however to remain backwards compatible in the event that users have customized their service to lock in a version of `wodby/xhprof` pre 3.x we should keep the original ENV variable in place.  ENV variables are cheap and I can see no harm in having a fallback.

JDDoesDev:
Fixes xhprof service to work with newer wodby/xhprof images.

closes #1507

### DIFF
--- a/stacks/overrides-xhprof.yml
+++ b/stacks/overrides-xhprof.yml
@@ -14,12 +14,11 @@ services:
 
   # XHProf Service to review
   xhprof:
-    image: ${XHPROF_IMAGE:-wodby/xhprof}
+    image: ${XHPROF_IMAGE:-wodby/xhprof:3.0.2}
     hostname: xhprof
     depends_on:
       - cli
     environment:
-      - PHP_XHPROF_OUTPUT_DIR=${XHPROF_OUTPUT_DIR:-/tmp/xhprof}
       - XHPROF_OUTPUT_DIR=${XHPROF_OUTPUT_DIR:-/tmp/xhprof}
     volumes:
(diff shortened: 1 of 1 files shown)


lmakarov (changes requested):
@JDDoesDev we should pin the `wodby/xhprof` version (`wodby/xhprof:3.0.2`) and then we don't need to keep the old variable name. The name was changed in the [wodby/xhprof:3.0.0 release](https://github.com/wodby/xhprof/releases/tag/3.0.0).

JDDoesDev:
@lmakarov that makes sense.  I'll make that change.

lmakarov (approved):

PR #1512, opened 2021-06-30
Show how to use the versions listed from the registry.

danshumaker:
Small example on how to specify a solr image.


docs/content/service/other/apache-solr.md
```diff
@@ -83,6 +83,12 @@ Legend:
 
 - `<image-repo>:<software-version>[-<image-stability-tag>][-<flavor>]`
 
+To use one of these versions put this variable in your .docksal/docksal.env file. For example to use the latest version put this in your docksal.env file
```
lmakarov, line 86: We use `fin config set` is the preferred way for settings a project configuration variable. Same result, but faster.

Take a look at https://docs.docksal.io/service/db/settings/#mysql-versions.

docs/content/service/other/apache-solr.md
```diff
@@ -83,6 +83,12 @@ Legend:
...
+
+```bash
+SOLR_IMAGE='docksal/solr:latest'
```
lmakarov, line 89: It's recommended to pid a specific version and almost never use `latest` as that WILL eventually result in unexpected changes.

lmakarov (changes requested):

lmakarov:
ping @danshumaker , @shelane

PR #1480, opened 2021-03-23
Issue #1476 Updates provider_pull_acquia_db()

acouch:
#1476

This updates `fin pull db` for Acquia.

This removes the code that checks for the latest backup since acquiacli does this with it's backup command: https://github.com/typhonius/acquia_cli/blob/master/src/Commands/DbBackupCommand.php#L183

This requires adding a new variable `ACQUIA_REALM` for ACP users. Would of course update the docs if this is pulled in.

Issue #1476, opened 2021-03-19
Update db pull for Acquia to use acquiacli


acouch:
**Description**

Acquia no longer supports ``ac-database-instance-backup-list`` since their Cloud Platform API version 1 reached EOL on 7/1/20.  I'd propose replacing  ``ac-database-instance-backup-list`` with ``acquiacli database:backup:download`` since it is already part of the default CLI.

I'd be happy to create a PR for this if the maintainers agree it is a good idea.

lmakarov:
> I'd be happy to create a PR for this if the maintainers agree it is a good idea.

@acouch Yes, absolutely. PRs are certainly welcome =)

@sean-e-dietrich maintains the `pull` subset of commands and can work with you on a PR.

PR #1521, opened 2021-07-16
Fix Acquia CLI auto-login

lpeabody:
Updates the Acquia CLI environment variables injected via services.yml so that they match the expected variables in startup.sh in docksal/cli https://github.com/docksal/service-cli/blob/develop/7.4/startup.sh#L140-L153.


shelane:
It needs to be updated in bin/fin as well.

shelane:
The variables are no longer used:
SECRET_ACAPI_EMAIL 
SECRET_ACAPI_KEY

lpeabody:
The Acquia provider pull method is a little more complicated. I have never personally used it, and looking through `fin` There is an implication that the Acquia provider pull method has a hard dependency on typhonius/acquia_cli:

```
# Check to see if acquia keys have been set.
provider_pull_acquia_precheck ()
{
	if [[ -z "${SECRET_ACAPI_EMAIL}" ]] || [[ -z "${SECRET_ACAPI_KEY}" ]]; then
		echo-error "SECRET_ACAPI_EMAIL and SECRET_ACAPI_KEY must be set in order to work with Acquia."
		exit 1;
	fi

	if [[ "${ARGV[0]}" == "init" ]]; then
		return;
	fi

	provider_pull_precheck
}
```

`SECRET_ACAPI_EMAIL` and `SECRET_ACAPI_KEY` are only used by acquia_cli, and that is removed from current images... so I think Acquia provider pull is likely broken, currently?

shelane:
@sean-e-dietrich Needs to comment on that.

lmakarov:
Related: #1476 / #1480. May need to be reviewed and merged together.

JDDoesDev:
@lpeabody 
> The Acquia provider pull method is a little more complicated. I have never personally used it, and looking through fin There is an implication that the Acquia provider pull method has a hard dependency on typhonius/acquia_cli:

At some point, `typhonius/acquia_cli` was phased out and is now handled by Acquia. I believe this was mentioned in #1480.  

Since that is the case, we should probably replace any `typhonius/acquia_cli` references with the new version and update any scripted commands to use the new syntax where applicable.

I have not been able to use `fin pull` with Acquia yet because the commands aren't quite where they need to be, example `fin acquiacli <command>` as mentioned in [the docs](https://docs.docksal.io/tools/acquia-cli/) vs actually needing to use `fin exec acli <command>`.  I'm assuming that once that gets straightened out, we should be able to hook into `fin pull` a little easier.

sean-e-dietrich:
I believe we should document the change of the variables and indicate how to include the old variables for versions of cli container

lmakarov:
Is this still relevant?

shelane:
The login stuff and the variables do need to be fixed. The pull changes is now moot. Do you want to close this and I can do a separate PR for the variables and login?

PR #1577, opened 2021-09-28
Release 1.16.2

lmakarov:
### New software versions

- fin v1.107.1

### Changes and improvements

- Fixed terminal colors in `fin` broken on some systems after update to v1.16.1


shelane (approved):
I tested the develop version on a system that was affected by the change in 1.16.1 and it works.

PR #1604, opened 2021-12-06
Update system-dns.md

marvoey:
With DOCKSAL_NO_DNS_RESOLVER=1 docksal will still try to start up the DNS service.
Need to use DOCKSAL_DNS_DISABLED=1 to actually disable the service completely.

### DIFF
--- a/docs/content/core/system-dns.md
+++ b/docs/content/core/system-dns.md
@@ -28,7 +28,7 @@ If you run into issues with DNS resolution, try disabling the automatic resolver
 
 ```bash
 fin system stop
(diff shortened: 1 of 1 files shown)


lmakarov (approved):

lmakarov:
Yeah, makes sense.

`DOCKSAL_NO_DNS_RESOLVER` was only supposed to disable the DNS configuration overrides on the host, but not the `docksal-dns` service. 

`DOCKSAL_DNS_DISABLED` is a newer flag and disables both for good.

PR #1628, opened 2022-02-09
Change cli debugging instructions

froboy:
In attempting to configure Xdebug with PHPStorm to debug with Drush I found that the existing instructions didn't correctly set the `PHP_IDE_CONFIG` variable, I think because `VIRTUAL_HOST` is in a different context than `docksal-local.env` so it's not able to be referenced.

Using the old instructions, this is what I get after `fin up`:

```
❯ fin exec printenv PHP_IDE_CONFIG
serverName=
```

My project is using the default `VIRTUAL_HOST` (as I'd imagine many are) and not setting it in `docksal.env`. After setting `PHP_IDE_CONFIG` in `docksal.yml`:

```
❯ fin exec printenv PHP_IDE_CONFIG
"serverName=my-project.docksal"
```

### DIFF
--- a/docs/content/tools/xdebug.md
+++ b/docs/content/tools/xdebug.md
@@ -77,11 +77,13 @@ First, follow [automatic](#phpstorm-automatic) or [manual](#phpstorm-manual) ins
 mapping settings in PHPStorm.
 
 To debug PHP CLI scripts, we have to tell PHPStorm which **existing** server configuration to use via the
-`PHP_IDE_CONFIG` variable. This can be done using the following commands:
+`PHP_IDE_CONFIG` variable. This must be added to `docksal.yml`:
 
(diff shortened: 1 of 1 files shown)


froboy:
Removed quotes as per https://stackoverflow.com/a/50176116/2566038.

sean-e-dietrich:
@froboy the documentation currently seems valid without the changes.

The `PHP_IDE_CONFIG` variable is currently present in the `cli` config https://github.com/docksal/docksal/blob/develop/stacks/services.yml#L148

I'm wondering if the wondering may need to be just changed a little bit to make it a little more apparent that you would need to add a `PHP_IDE_CONFIG` to debug cli

shelane (approved):

PR #1664, opened 2022-04-19
Update terminus.md

halisonfernandes:


shelane (changes requested):
We really should instead replace it with: 

Follow the [official docs](https://pantheon.io/docs/terminus/install) to generate your machine token and then add the token to your global configuration:
```
fin config set --global SECRET_TERMINUS_TOKEN=XMAG92S9_9gf5Mlhb7-JIEjVwYYhY-MGuKcspAdL0CjkU
```

lmakarov:
@shelane Absolutely agree. There are also a few more places where this need to happen.
Created a separate issue to address this - https://github.com/docksal/docksal/issues/1665

I think we can accept this particular PR as-is and give @halisonfernandes a contribution credit :)

PR #1704, opened 2022-08-12
Update multiple-web.md

khaledzaidan:
Mention that user can add "Include conf/extra/includes/host.conf" to their vhost to avoid writing it all from scratch.


lmakarov (approved):

lmakarov:
Thanks!

PR #1720, opened 2022-10-20
Fix checking web container in project

matthiasseghers:
Fin share checks if a web container is running inside the project. The check looks for "Up" but "fin ps web" returns "running", therefore reporting the project as not having a running web container.


lmakarov (approved):

lmakarov:
Thanks!

PR #1799, opened 2023-10-24
Fix bash completion

### LINKED ISSUE #1798: Bash autocomplete could not be installed
**Description**

1. The `fin update --bash-complete` could not install the bash autocomplete, because it needs root permissions to create/write `/usr/local/bin/fin-bash-autocomplete`. 
2. The documentation is missing.
3. The autocomplete doesn't discover grouped commands (ex: `fin myproject/init`) 

1. Execute `fin update --bash-complete` or ` sudo fin update --bash-complete`, both fail, first one due to permission on `/usr/local/bin/fin-bash-autocomplete`, the second one because it updates the root `.bash_profile`.

obriat:
fixes #1798

### DIFF
--- a/bin/fin
+++ b/bin/fin
@@ -4973,7 +4973,7 @@ update_config_files ()
 install_bash_autocomplete ()
 {
 	local destination="$FIN_AUTOCOMPLETE_PATH"
-	tee "$destination" >/dev/null <<'EOF'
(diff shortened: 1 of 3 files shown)


lmakarov (approved):
Thanks for your contribution!

I'm going to approve and  merge as-is, then make some adjustments.

PR #1820, opened 2024-04-01
replaces mailhog service with mailpit

shelane:
This was pointed out in the addons repo:

https://github.com/docksal/addons/issues/93

Here is the mailpit repo and the latest tag is 1.15.1, which was released Mar 30, 2024.
https://github.com/axllent/mailpit


shelane:
This includes a change in the `MAILHOG_IMAGE` variable to `MAILPIT_IMAGE`, which could be considered a backwards breaking change. We could keep the image name for the next release of Docksal and change it only on a major version. What do you think?

sean-e-dietrich (approved):

lmakarov:
> This includes a change in the MAILHOG_IMAGE variable to MAILPIT_IMAGE, which could be considered a backwards breaking change.

This won't cause a breaking change, so should be OK. Let's proceed with this swap. Thanks!

lmakarov (approved):

PR #1843, opened 2025-01-21
Enhance compatibility for Docker Desktop alternatives on macOS

syntlyx:
- Updated the `docker_desktop_version` function to return a static version (4.37.2) if `/Applications/Docker.app` does not exist, allowing better support for alternatives like Colima and OrbStack.
- Modified the logic for `overrides_dd_bind_file` to skip applying overrides if the system is macOS and `/Applications/Docker.app` is absent, ensuring seamless behavior with alternative setups.
- Enhanced the `configure_mounts` function to skip NFS configuration on macOS unless `DOCKSAL_VOLUMES` is set to `nfs`, aligning with use cases where NFS is not required or supported by alternatives.
- These changes provide improved compatibility and flexibility for users relying on Docker Desktop alternatives on macOS.


banesbyte:
@lmakarov We are using Colima or OrbStack, and these changes make it easier to switch from Docker Machine. Additionally, since most people use bind and NFS is not in use, it would be best to avoid starting NFS shares at user preference. Let me know if you want us to improve the requested changes or it somehow affects others.

PR #1848, opened 2025-03-28
Update acquia-cli.md

D528lgym:


shelane:
What is this change for and why? This page is still for the Acquia CLI.

